PR #6, opened 2019-06-17
Feature/tao 8246/add form components/widget

jsconan:

Requires: 

So here's what we have here:
- `ui/form/widget/widget` is the factory that will be used to create the widgets. It implements a providers registry, and create a component that exposes a common API. Then each widget will be a provider to register in this registry and will have to implement the specific parts.
- `ui/form/widget/providers/default` is the default internal widget implementation, and is applied each time the given provider does not implement the invoked API.

Unit tests:
```
npm i
npm run test form
```

```
npm run test:keepAlive
```
Then open a browser on http://127.0.0.1:8082/test/form/widget/widget/test.html


jsconan:
Added the fix for bundling under Windows so that you can now run the tests.

src/form/widget/widget.js
```diff
@@ -0,0 +1,379 @@
...
+            delegate('reset');
+            setInvalidState(false);
+            return this;
+        },
+
+        /**
+         * Serializes the value of the widget
+         * @returns {widgetValue}
+         */
+        serialize() {
```
krampstudio: I don't like the `serialize` name, it is misleading.This method role is to get the widget values and should have been called  `getValues`, no ? 
You don't need to change it but this can be an improvement for the next time.

src/form/widget/widget.js
```diff
@@ -0,0 +1,379 @@
...
+                    return Promise.reject(err);
+                });
+        },
+
+        /**
+         * Triggers the change event
+         * @returns {widgetForm}
+         * @fires change
+         */
+        notify() {
```
krampstudio: I also found this method name too broad : `change` or `update` ? What do you think ?

src/form/widget/widget.js
```diff
@@ -0,0 +1,379 @@
...
+             */
+            this.trigger('change', this.getValue(), this.getUri());
+            return this;
+        },
+
+        /**
+         * Gets access to the actual form element
+         * @returns {jQuery|null}
+         */
+        getWidgetElement() {
```
krampstudio: `getElement` would also be more correct

test/form/widget/widget/test.html
```diff
@@ -0,0 +1,73 @@
...
+        <label>
+            <span>Changes:</span>
+            <textarea class="change-output"></textarea>
+        </label>
+    </div>
+</div>
+</body>
+</html>
+
+
```
krampstudio: could you please remove the  blank lines at the end

src/form/widget/widget.js
krampstudio: but it returns already the component's element. Forget about it

src/form/widget/widget.js
jsconan: What about `notifyChange` ?

jsconan:
If you don't mind, I will apply the method names change in the PR for the form as some are used in several places

src/form/widget/widget.js
jsconan: I will replace it by `serializeValue()`, will do the change in #17

src/form/widget/widget.js
jsconan: I will remove this API, it should remain internal. I though some widget implementation could use it, but this should not be the case.

jsconan:
Finally I managed to do the changes in this PR, I will cherry pick the commit in other PRs

PR #11, opened 2019-06-18
Feature/tao 8246/add form components/widget hiddenbox

jsconan:

Requires: 

So here's what we have here:
- `ui/form/widget/hiddenBox` is a double password input widget

Unit tests:
```
npm i
npm run test form
```

```
npm run test:keepAlive
```
Then open a browser on http://127.0.0.1:8082/test/form/widget/hiddenBox/test.html

### DIFF
--- a/src/form/widget/providers/hiddenBox.js
+++ b/src/form/widget/providers/hiddenBox.js
@@ -1,0 +1,145 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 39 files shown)


test/form/widget/hiddenBox/test.js
```diff
@@ -0,0 +1,821 @@
...
+            .on('init', function () {
+                assert.equal(this, instance, 'The instance has been initialized');
+            })
+            .on('ready', function () {
+                assert.equal($container.children().length, 2, 'The container contains an element');
+                ready();
+            })
+            .on('change', function (value, uri) {
+                this.validate();
+                $outputChange.val('value of [' + uri + '] changed to "' + value + '"\n' + $outputChange.val());
```
ricproenca-tao: Consider replace `value` for `JSON.stringify(value)` once value is an object.

jsconan: Done

src/form/widget/providers/hiddenBox.js
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * Initialize the widget.
+     * @param {hiddenBoxConfig} config
+     */
+    init(config) {
+        // the type will be reflected to the HTML markup
+        config.widgetType = 'hidden-box';
+
+        // config for the confirmation field
+        config.confirmation = _.defaults(config.confirmation || {}, {
```
krampstudio: Maybe the confirmation needs to be optional.
It can be also for a future improvement

jsconan: Do you mean the widget could be either single password input or confirmed password input?
Should not it be better then to have another component that only supplies a single password?

PR #12, opened 2019-06-18
Feature/tao 8246/add form components/widget radiobox

jsconan:

Requires: 

So here's what we have here:
- `ui/form/widget/radioBox` is multi radio buttons widget

Unit tests:
```
npm i
npm run test form
```

```
npm run test:keepAlive
```
Then open a browser on http://127.0.0.1:8082/test/form/widget/radioBox/test.html

### DIFF
--- a/src/form/widget/providers/radioBox.js
+++ b/src/form/widget/providers/radioBox.js
@@ -1,0 +1,99 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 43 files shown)


src/form/widget/providers/radioBox.js
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Gets the value of the widget
+     * @returns {String}
+     */
+    getValue() {
+        let value = this.getConfig().value || '';
+
+        if (this.is('rendered')) {
+            value = this.getElement()
+                .find(`.option input[name="${this.getUri()}"]:checked`)
```
krampstudio: are you going to have multiple radios ? 
 `.option :checked` should be sufficient (like for the checkboxes) or is there particular reason to have the URI in the selector here ?

src/form/widget/providers/radioBox.js
```diff
@@ -0,0 +1,99 @@
...
+                .val() || '';
+        }
+
+        return value;
+    },
+
+    /**
+     * Sets the value of the widget
+     * @param {String} value
```
krampstudio: do you expect the URI from the range ?

src/form/widget/providers/radioBox.js
jsconan: Good catch!

src/form/widget/providers/radioBox.js
jsconan: Reworded the JSDoc

PR #17, opened 2019-06-18
Feature/tao 8246/add form components/form

jsconan:

Requires: 
 - [x] https://github.com/oat-sa/tao-core-ui-fe/pull/7
 - [x] https://github.com/oat-sa/tao-core-ui-fe/pull/8
 - [x] https://github.com/oat-sa/tao-core-ui-fe/pull/9
 - [x] https://github.com/oat-sa/tao-core-ui-fe/pull/10
 - [x] https://github.com/oat-sa/tao-core-ui-fe/pull/11
 - [x] https://github.com/oat-sa/tao-core-ui-fe/pull/12
 - [x] https://github.com/oat-sa/tao-core-ui-fe/pull/16

So here's what we have here:
- `ui/form/form` is the main component, handling the form and exposing an API. By default the component does not contain any widgets or buttons. The API allows to set widgets and buttons.

Unit tests:
```
npm i
npm run test form
```

```
npm run test:keepAlive
```
Then open a browser on http://127.0.0.1:8082/test/form/form/test.html


src/form/form.js
```diff
@@ -0,0 +1,669 @@
...
+                .then(() => new Promise(resolve => {
+                    const button = buttonFactory(definition);
+                    buttons.set(definition.id, button);
+                    button
+                        .on('click.form', () => {
+                            /**
+                             * @event button
+                             * @param {String} id
+                             */
+                            this.trigger('button', definition.id);
```
krampstudio: Is it possible to not trigger an event called `button` but directly the unlerying action : `publish`, `submit`, `reset` or `clear` ?

src/form/form.js
```diff
@@ -0,0 +1,669 @@
...
+                                 * @param {String} uri
+                                 * @param {String} value
+                                 */
+                                this.trigger('change', definition.uri, value);
+
+                                /**
+                                 * @event change-<uri>
+                                 * @param {String} value
+                                 */
+                                this.trigger(`change-${definition.uri}`, value);
```
krampstudio: what's the purpose of this event ?

src/form/form.js
```diff
@@ -0,0 +1,669 @@
...
+                     * @param {widgetValue[]} values
+                     */
+                    this.trigger('submit', this.serialize());
+                })
+                .catch(reason => {
+                    /**
+                     * @event invalid
+                     * @param {Object} reason
+                     */
+                    this.trigger('invalid', reason);
```
krampstudio: Since the rejection is used to handle user error, it would be good to also support system errors.
If  `reason` is an `Error` (something went wrong within the validation system) we should trigger the `error` event

src/form/form.js
```diff
@@ -0,0 +1,669 @@
...
+ * @property {Object} [ranges] - An optional list of ranges for the widgets (@see widgetConfig.range)
+ */
+
+/**
+ * Some default config
+ * @type {formConfig}
+ */
+const defaults = {
+    formAction: '#',
+    formMethod: 'get'
```
krampstudio: I'd say no action nor method by default

src/form/tpl/form.tpl
```diff
@@ -0,0 +1,7 @@
+<div class="form-component">
+    <h2 class="form-title">{{title}}</h2>
+    <form action="{{formAction}}" method="{{formMethod}}">
```
krampstudio: also no action nor method by default

src/form/form.js
krampstudio: or what about a definition that contains the action : 
```js
return formFactory(container, {
  buttons : [{
        type: 'info',
        id: 'submit',
        label: config.submitText,
        icon: config.submitIcon,
        action : form => form.submit()
), {
      type: 'neutral',
      id: 'reset',
      label: config.resetText,
      icon: config.resetIcon
      action : form => form.reset()
), {
      type: 'foo',
      id: 'reset',
      label: config.resetText,
      icon: config.resetIcon
      action : form => {
          form.trigger('foo')
      }
}])
```

jsconan: Well the idea was to avoid conflict with other events, hence the namespace.
So to listen to any button click, the `button` event is available. For a particular button, the event `button-<id of the button>` is available.
Having action built with the button definition could be also an option, but it will reduce the possibilities, as only one listener will be applicable, and no event pipeline will be available too.

src/form/form.js
jsconan: The purpose is to give the ability to listen change on a particular widget.

src/form/form.js
jsconan: done

PR #23, opened 2019-06-19
Fix/nex 147 windows separator fix for local externals

btamas:
Jira: https://oat-sa.atlassian.net/browse/NEX-147

Test on Windows:
`npm run prepare`
`npm run test`


btamas:
> In fact it was working fine for a subset, but unfortunately it seems some unit tests are still failing.
> Maybe should you apply the same patch on every place where a FS path is in use.

It was just a helper in `rollup.config.js`. I fixed it and I increased the timeout in `themeLoader` test, because 50ms was not always enough.

PR #24, opened 2019-06-20
Feature/nex 149 add babel setup

krampstudio:
Ensure the sources can be written according the JavaScript coding guide in ES2015 and the generated AMD modules are compatible in `dist`.
This PR setup Babel to transpile the modules

How to test : 
```
npm i
npm run build
npm test
```
The file `src/component.js` has received minor changes to prove the transpilation works, so you can verify it in `dist`

### DIFF
--- a/build/rollup.config.js
+++ b/build/rollup.config.js
@@ -124,7 +124,6 @@ export default inputs.map(input => {
             babel({
                 presets: [[
                     '@babel/env', {
-                        targets: 'extends @oat-sa/browserslist-config-tao',
                         useBuiltIns: false
                     }
                 ]]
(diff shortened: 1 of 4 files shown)


build/rollup.config.js
```diff
@@ -119,7 +120,16 @@ export default inputs.map(input => {
                         );
                     }
                 }
-            }
+            },
+            babel({
+                presets: [[
+                    '@babel/env', {
+                        targets: 'extends @oat-sa/browserslist-config-tao',
```
btamas: It looks redundant for me, because `browserlist` is defined in `package.json`.

PR #27, opened 2019-06-24
Add build watcher

jsconan:
Add a CLI command to build and watch for changes: `npm run build:watch`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,13 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "0.16.1",
+    "version": "0.16.2",
     "displayName": "TAO Core UI",
     "description": "UI libraries of TAO",
     "scripts": {
         "test:keepAlive": "node ./build/webserver.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,12 +1,13 @@
...
     "name": "@oat-sa/tao-core-ui",
-    "version": "0.16.1",
+    "version": "0.16.2",
     "displayName": "TAO Core UI",
     "description": "UI libraries of TAO",
     "scripts": {
         "test:keepAlive": "node ./build/webserver.js",
         "test": "node ./build/testrunner.js",
         "build": "rollup --config ./build/rollup.config.js",
+        "build:watch": "rollup --config ./build/rollup.config.js --watch",
```
btamas: Do not forget to update `README.md` with the new command.

jsconan: Ok, changes made.
However, only the change in CSS are watched, we still need to run `npm run buildScss` to apply changes from SCSS. It will require some extra work top make the `buildScss` watchable

PR #28, opened 2019-06-24
Fix/tao 8518 russian font

anshul119:
Ticket: https://oat-sa.atlassian.net/browse/TAO-8518

What changed?

I changed the regular and heading font family to source sans pro from AppleGothic because AppleGothic does not support Russian Alphabets.

Related PR: 

- [ ] https://github.com/oat-sa/tao-core/pull/2193

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ typings/
 /node_modules
 /dist
 css/
+
+# phpstorm files
+.idea/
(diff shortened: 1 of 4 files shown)


btamas:
I think it will affect all project. Is that approved?
BTW, I think it is a breaking change.

anshul119:
> I think it will affect all project. Is that approved?
> BTW, I think it is a breaking change.

Yes, it has been approved by UX

anshul119:
@btamas So do you think we should upgrade the major version of the package? I feel it's not as impactful that we need to change the major version. What do you think @krampstudio ?

PR #29, opened 2019-06-25
[WIP] Feature/tao 8246/use new ui form in resourceselector

jsconan:

A side PR to replace the use of `ui/generis/form` by `ui/form/form`.
The only place where it was in use was in the `resourceSelector`.

The biggest change, apart the change of component, is that the new form component does not render synchronously, so the consumer should listen to the `ready` event to be able to properly interact with it. For that reason, the `filters.update()` method now emits an `update` event when the form filter is ready. It also emits a `ready` event when the component has been fully rendered.

To test the PR:
```
npm run prepare
npm run test
```

To visually interact with the component:
```
npm run test:keepAlive
```
Then open your browser on http://127.0.0.1:8082/test/resource/filters/test.html

### DIFF
--- a/src/resource/filters.js
+++ b/src/resource/filters.js
@@ -25,6 +25,7 @@
 
 import _ from 'lodash';
 import __ from 'i18n';
+import promiseTimeout from 'core/promiseTimeout';
 import component from 'ui/component';
 import formFactory from 'ui/form/simpleForm';
 import widgetDefinitions from 'ui/form/widget/definitions';
@@ -103,7 +104,7 @@ export default function filtersFactory($container, config) {
 
             /**
              * Reset the filter form
-             * @return {filter} chains
+             * @return {Promise}
              */
             reset() {
                 return this.update(this.config.data);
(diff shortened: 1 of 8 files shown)


src/resource/filters.js
```diff
@@ -194,15 +198,38 @@ export default function filtersFactory($container, config) {
             this.render($container);
         })
         .on('render', function() {
-            if (this.config.data) {
-                this.update(this.config.data);
-            }
+            Promise.race([
+                new Promise(resolve => {
+                    if (this.config.data) {
+                        this
```
krampstudio: I'm wondering if the  `update` method shouldn't return a Promise (in addition to the event) this would make the workflow more straight forward ?

src/resource/filters.js
```diff
@@ -194,15 +198,38 @@ export default function filtersFactory($container, config) {
...
+                            .on('update.ready', () => {
+                                this.off('update.ready');
+                                resolve();
+                            })
+                            .update(this.config.data);
+                    } else {
+                        resolve();
+                    }
+                }),
+                new Promise((resolve, reject) => {
```
krampstudio: As a nice to have, what about extracting this pattern to something like `core/promiseTimeout`

jsconan: Agree, here is the PR:
https://github.com/oat-sa/tao-core-sdk-fe/pull/21

src/resource/filters.js
jsconan: Yes that would be better. That will also probably requires some rework on the consumer. I will try that approach.

jsconan:
The filter now makes use of `core/promiseTimeout`, and the API `filter.update()` returns a Promise.
However, I still need to propagate the API change in the `resourceSelector`, to ensure all works smoothly. For now it seems ok as the unit tests don't fail, but more tests would be needed to validate the change.
That means the PR is not complete yet.

jsconan:
Stalled PR

PR #32, opened 2019-06-26
Feature/tao 8559/message and icon in bulkActionPopup

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-8559

Customized bulkActionPopup with message and icon
image

Also:
Fixed the modal close icon (button) styles
Added src folder to eslint check
 
How to test:
- `npm run build`
- `npm run test:keepAlive`
- go to http://127.0.0.1:8082/test/bulkActionPopup/test.html

Related PR
https://github.com/oat-sa/tao-core/pull/2196

### DIFF
--- a/src/bulkActionPopup.js
+++ b/src/bulkActionPopup.js
@@ -65,7 +65,7 @@ export default function bulkActionPopupFactory(config) {
          * Validates the dialog, and closes it (action performed when hitting the Ok button)
          * @returns {Boolean} Returns `true` if the dialog has been successively validated (and closed)
          */
-        validate: function validate() {
+        validate() {
             const $element = this.getElement();
 
             if ($element) {
@@ -85,7 +85,7 @@ export default function bulkActionPopupFactory(config) {
         /**
          * Cancels and closes the dialog
          */
-        cancel: function cancel() {
+        cancel() {
             this.trigger('cancel');
             this.destroy();
         }
@@ -123,7 +123,6 @@ export default function bulkActionPopupFactory(config) {
 
         // renders the component
         .on('render', function() {
-            const self = this;
             const $element = this.getElement();
 
             initModal({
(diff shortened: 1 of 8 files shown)


src/bulkActionPopup/scss/bulkActionPopup.scss
```diff
@@ -69,4 +51,14 @@
     .multiple .resource-count{
         font-weight : bold;
     }
+
+    .message{
+        @include flex-container();
+        @include vendor-prefix(align-items, center, property);
```
jsconan: This repo is using cssprefixer, no need to prefix anymore :)

src/bulkActionPopup.js
```diff
@@ -122,17 +123,16 @@ export default function bulkActionPopupFactory(config) {
 
         // renders the component
         .on('render', function() {
-            var self = this;
-            var $element = this.getElement();
-            var $reason;
+            const self = this;
```
jsconan: While moving to ES6, it could be good too to also get rid of lexical scope enforcer (using arrow functions instead when the lexical scope should be the parent one)
Also using shorthand notation for object members: `validate: function validate() { ... }`  => `validate() { ... }`.
Those changes are not heavy and won't change the behavior, so we should apply them when we touch a file and it is not too big.

src/bulkActionPopup/scss/bulkActionPopup.scss
bziondik: fixed

src/bulkActionPopup.js
bziondik: changed to arrow functions and shorthand notation

src/bulkActionPopup.js
```diff
@@ -122,40 +123,38 @@ export default function bulkActionPopupFactory(config) {
...
                         // ensure the keyboard navigation is taking care of the possible new fields
                         initNavigator();
                     }
-                    self.trigger('change', state);
+                    this.trigger('change', state);
                 })
-                .on(namespaceHelper.namespaceAll('change', _ns), 'textarea', function() {
-                    state.comment = $(this).val();
-                    self.trigger('change', state);
+                .on(namespaceHelper.namespaceAll('change', _ns), 'textarea', (e) => {
```
jsconan: our code style says "no parenthesis around single parameter for arrow functions"

bziondik: @jsconan We don't have this rule in the Style guide https://oat-sa.atlassian.net/wiki/spaces/DEV/pages/79037064/TAO+Frontend+Coding+guide
and in .eslintrc.json
Should I add it in .eslintrc.json ?
`"arrow-parens" :                ["error", "as-needed"]`
https://eslint.org/docs/rules/arrow-parens

jsconan: Ok, no worries. We use to apply the rule, but it is not expressed in the coding guide.
This is not a blocker, so I will merge the PR

PR #37, opened 2019-07-10
Fix/tao 8122/loading bar sync

ricproenca-tao:
**Related to task:** 
https://oat-sa.atlassian.net/browse/TAO-8122

**Description:**
Loading bar out of sync.
The loading bar stops before the new items are fully rendered in the table rows.

With this fix, we pretend to see the new items to be added in the datatable at the same time that the loading bar stops.
Also, 

**Requested for:**
- [ ] https://github.com/oat-sa/extension-tao-nccer-delivery/pull/926

### DIFF
--- a/src/datatable.js
+++ b/src/datatable.js
@@ -190,6 +190,8 @@ var dataTable = {
      */
     _refresh: function($elt, data) {
         // TODO: refresh only rows with data, not all component
+
+        loadingBar.start();
         if (data) {
             this._render($elt, data);
         } else {
@@ -210,8 +212,6 @@ var dataTable = {
         var options = $elt.data(dataNs);
         var parameters;
         var ajaxConfig;
-
-        loadingBar.start();
 
         if (!$filter) {
             $filter = $('.filter', $elt);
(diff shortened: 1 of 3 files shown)


src/datatable.js
```diff
@@ -211,8 +213,6 @@ var dataTable = {
         var parameters;
         var ajaxConfig;
 
-        loadingBar.start();
```
jsconan: well, I guess this should stay: `_query()` could called outside of `_refresh()`

ricproenca-tao: Yes, you're right.
Also, has @btamas pointed out, `_query` is also called from other methods...

ricproenca-tao:
@jsconan 
As for the unit test, I don't think we need it. 
This fix only changes loading bar timings, doesn't change behavior.

src/datatable.js
```diff
@@ -212,7 +212,7 @@ var dataTable = {
         var ajaxConfig;
 
         loadingBar.start();
-
+
```
btamas: Just a useless tab.

PR #39, opened 2019-07-12
Fix/tao 8663/Set proper button context in datatable

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-8663

The `datatable` can manage a list of actions on each row. Unfortunately, the context of each action does not allow to apply dynamic behavior related to the row data while rendering, like for instance display or hide the button if the row contains the expected data.

With this fix, each action can reflect the row. For instance the label of the button can be set accordingly or the button can be hidden.

Example:
```javascript
$('.table').datatable({
    url: '/demo-data/actions',
    rows: 10,
    actions: [
        {
            id: 'remove',
            label() {
                if (this.login && this.login.length) {
                    return 'Delete account';
                }
                return 'Remove user';
            },
            icon: 'bin',
            action(id, row) { ... }
        },
        {
            id: 'connect',
            label: 'Connect',
            icon: 'user',
            hidden() {
                return !this.login || !this.login.length;
            },
            action(id, row) { ... }
        },
        {
            id: 'create',
            label: 'Create account',
            icon: 'user',
            hidden() {
                return this.login && this.login.length;
            },
            action(id, row) { ... }
        }
    ],
    model: [
        {
            id: 'login',
            label: 'Login'
        },
        {
            id: 'login',
            label: 'Login'
        },
        {
            id: 'email',
            label: 'Email'
        },
        {
            id: 'firstname',
            label: 'First Name'
        },
        {
            id: 'lastname',
            label: 'Last Name'
        },
        {
            id: 'phone',
            label: 'Phone'
        },
        {
            id: 'country',
            label: 'country'
        }
    ]
});
```

The unit tests should not fail
```
npm run prepare
npm run test
```

To see the feature in action:
```
npm run build
npm run test:keeAlive
```
Then open a browser on http://127.0.0.1:8082/test/datatable/demo/test.html

**Note**
- [ ] After merge, don't forget to release and publish to npm

### DIFF
--- a/src/datatable/tpl/layout.tpl
+++ b/src/datatable/tpl/layout.tpl
@@ -85,15 +85,7 @@
                                     {{#if id}}
                                         {{#with ../../../../this}}
                                             {{#unless ../hidden}}
-                                                {{#if ../../disabled}}
-                                                    {{#with ../../../this}}
-                                <button class="btn-info small {{id}}"{{#if title}} title="{{title}}"{{/if}} disabled="disabled">{{#if icon}}<span class="icon-{{icon}}"></span> {{/if}}{{#if label}} {{label}}{{/if}}</button>
-                                                    {{/with}}
-                                                {{else}}
-                                                    {{#with ../../../this}}
-                                <button class="btn-info small {{id}}"{{#if title}} title="{{title}}"{{/if}}>{{#if icon}}<span class="icon-{{icon}}"></span> {{/if}}{{#if label}} {{label}}{{/if}}</button>
-                                                    {{/with}}
-                                                {{/if}}
+                                <button class="btn-info small {{../../id}}"{{#if ../../title}} title="{{../../../title}}"{{/if}}{{#if ../../disabled}} disabled="disabled"{{/if}}>{{#if ../../icon}}<span class="icon-{{../../../icon}}"></span>{{/if}} {{../../label}}</button>
                                             {{/unless}}
                                         {{/with}}
                                     {{else}}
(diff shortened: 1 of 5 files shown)


test/datatable/demo/data.json
```diff
@@ -66,7 +66,7 @@
     "state": "Rhode Island",
     "street": "2492 Hermiston Lodge",
     "country": "Bangladesh",
-    "login": "Blaise.Dicki",
+    "login": "",
```
anshul119: Why did we remove all the names from this file?

test/datatable/demo/test.js
```diff
@@ -848,11 +848,54 @@ define([
...
+                            action: function(id, row) {
+                                var dialogConfig = {
+                                    message: 'Deletion is not supported yet!',
+                                    content: 'That would affect user id #' + id,
+                                    buttons: 'ok',
+                                    autoRender: true,
+                                    autoDestroy: true
+                                };
+                                if (row.login && row.login.length) {
+                                    dialogConfig.message = 'Account deletion is not supported yet!';
```
anshul119: Messages in both cases are similar. In this case, what purpose is this condition serving?

test/datatable/demo/test.js
```diff
@@ -889,6 +936,121 @@ define([
...
+                            label: 'Context',
+                            type: 'actions',
+                            actions: [
+                                {
+                                    id: 'file',
+                                    label: 'File',
+                                    icon: 'item',
+                                    action: function(id, row) {
+                                        dialog({
+                                            message: "Use's file #" + id,
```
anshul119: the apostrophe is not needed here

src/datatable/tpl/layout.tpl
```diff
@@ -85,15 +85,7 @@
...
-                                                {{#if ../../disabled}}
-                                                    {{#with ../../../this}}
-                                <button class="btn-info small {{id}}"{{#if title}} title="{{title}}"{{/if}} disabled="disabled">{{#if icon}}<span class="icon-{{icon}}"></span> {{/if}}{{#if label}} {{label}}{{/if}}</button>
-                                                    {{/with}}
-                                                {{else}}
-                                                    {{#with ../../../this}}
-                                <button class="btn-info small {{id}}"{{#if title}} title="{{title}}"{{/if}}>{{#if icon}}<span class="icon-{{icon}}"></span> {{/if}}{{#if label}} {{label}}{{/if}}</button>
-                                                    {{/with}}
-                                                {{/if}}
+                                <button class="btn-info small {{../../id}}"{{#if ../../title}} title="{{../../../title}}"{{/if}}{{#if ../../disabled}} disabled="disabled"{{/if}}>{{#if ../../icon}}<span class="icon-{{../../../icon}}"></span>{{/if}} {{../../label}}</button>
```
anshul119: Do we have a limit for line length? This seems too big to be on a single line and can be broken down in multiple to make it more readable :)

test/datatable/demo/data.json
jsconan: Well, it was to give some context for the visual playground: with a login, set a button to connect, without propose to create the account

test/datatable/demo/test.js
jsconan: This is a visual playground, in context of testing the feature

test/datatable/demo/test.js
jsconan: oops, copy&paste

src/datatable/tpl/layout.tpl
jsconan: Indeed, we could break in several lines. Usually in templates we tend to allow long lines, but anyway, I will apply some reformatting ;)

PR #46, opened 2019-07-25
Feature/nex 242/tabs component

oatymart:
https://oat-sa.atlassian.net/browse/NEX-242

Draft PR for early review.

- created a `ui/tabs` component which is initialised with an array of tab data
- `label` and `name` properties are mandatory for each tab (for now)
- optional auto-toggling of external panels connected to the tabs by data-attributes
- optional onClick properties for convenience of extra callbacks
- 3 different ways to programmatically select a tab
- mostly covered with unit tests
- minor functionality could change, based on experiences integrating into NEX-243

I suggest to run `npm run test:keepAlive` first, to experience the component in a visual playground.

### DIFF
--- a/src/tabs.js
+++ b/src/tabs.js
@@ -20,7 +20,7 @@
  */
 /**
  * @example
- * var instance = tabs({
+ * const instance = tabs({
  *   renderTo: $container,
  *   tabs: [
  *     { label: 'TAO Local', name: 'local-delivery' },
@@ -169,6 +169,21 @@ const tabsApi = {
         $(`[data-tab-content="${name}"]`).removeClass('hidden');
 
         this.trigger(`show-tab-content.${ns}`, name);
+    },
+
+    /**
+     * Gets the current active tab (if any)
+     * @returns {Object|null} name and index of tab
+     */
+    getActiveTab() {
+        const activeIndex = tabs.findIndex(t => t.active === true);
+        if (activeIndex === -1) {
+            return null;
+        }
+        return {
+            name: tabs[activeIndex].name,
+            index: activeIndex
+        };
     }
 };
 
--- a/src/tabs.js
+++ b/src/tabs.js
@@ -41,13 +41,11 @@ const ns = 'tabs';
 /**
  * Default config
  * @param {Integer} activeTabIndex - the index of the tab to start on
- * @param {Boolean} showHideTargets - if true, no need to wire up tabs to panels, it's automatic
- * @param {String} targetDataAttr - the data attribute suffix used to connect tabs to panels
+ * @param {Boolean} showHideTargets - if true, no need to wire up tabs to content, it's automatic
  */
 const tabsDefaults = {
     activeTabIndex: 0,
-    showHideTargets: true,
-    targetDataAttr: 'panel'
+    showHideTargets: true
 };
 
 /**
@@ -88,10 +86,9 @@ const tabsApi = {
      */
     connectTabs() {
         const $tabBar = this.getElement();
-        const targetDataAttr = this.config.targetDataAttr;
 
         for (let tab of tabs) {
-            $tabBar.find(`[data-controlled-${targetDataAttr}="${targetDataAttr}-${tab.name}"]`)
+            $tabBar.find(`[data-tab-name="${tab.name}"]`)
                 .off('click')
                 .on('click', () => {
                     this.activateTabByName(tab.name);
@@ -121,7 +118,7 @@ const tabsApi = {
 
     /**
      * Activates a single tab (deactivating others)
-     * Triggers the automatic showing & hiding of target panels
+     * Triggers the automatic showing & hiding of target tab-contents
      * Triggers onClick functions of the tabs
      * @param {Number} index - zero-based
      * @param {Boolean} [callOnClick=true] - if false, skips the onClick call
@@ -131,7 +128,6 @@ const tabsApi = {
      */
     activateTabByIndex(index, callOnClick = true) {
         const $tabBar = this.getElement();
-        const targetDataAttr = this.config.targetDataAttr;
 
         if (typeof index !== 'number' || index < 0 || index >= tabs.length) {
             throw new TypeError(`No tab exists at index: ${index}`);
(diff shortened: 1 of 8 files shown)


src/tabs.js
```diff
@@ -0,0 +1,209 @@
...
+     * Triggers onClick functions of the tabs
+     * @param {Number} index - zero-based
+     * @param {Boolean} [callOnClick=true] - if false, skips the onClick call
+     * @returns {tabs} instance
+     * @fires activate-tab
+     * @throws {TypeError} on invalid index param
+     */
+    activateTabByIndex(index, callOnClick = true) {
+        const $tabBar = this.getElement();
+        const attr = this.config.targetDataAttr;
```
ferenckiss89: pls rename attr to something more descriptive

src/tabs.js
```diff
@@ -0,0 +1,209 @@
...
+        return this;
+    },
+
+
+    /**
+     * Activates a single tab by its name
+     * (pass-through method to activateTabByIndex)
+     * @param {String} name - human-readable identifier
+     * @param {Boolean} [callOnClick=true] - if false, skips the onClick call
+     * @returns {tabs} instance
```
ferenckiss89: returns with the instance or nothing :)

src/tabs.js
```diff
@@ -0,0 +1,209 @@
...
+     * @fires show-panel
+     */
+    showDataPanel(name) {
+        const attr = this.config.targetDataAttr;
+
+        if (!name || name.length === 0) {
+            return;
+        }
+        $(`[data-${attr}]`).addClass('hidden');
+        var $temp = $(`[data-${attr}="${attr}-${name}"]`);
```
ferenckiss89: const, or even variable is not needed

test/tabs/test.js
```diff
@@ -0,0 +1,386 @@
...
+            typeof instance[data.name],
+            'function',
+            'The tabs instance exposes a "' + data.name + '" function'
+        );
+        instance.destroy();
+    });
+
+    QUnit.test('init', function(assert) {
+        var config = {
+        };
```
ferenckiss89: extra enter not needed

test/tabs/test.js
```diff
@@ -0,0 +1,386 @@
...
+        assert.ok($tabsDom.find('li.tab').eq(0).hasClass('active'), 'First tab is activated');
+        assert.notOk($tabsDom.find('li.tab').eq(1).hasClass('active'), 'Second tab is deactivated');
+        assert.notOk($tabsDom.find('li.tab').eq(2).hasClass('active'), 'Third tab is deactivated');
+
+        instance.destroy();
+    });
+
+    QUnit.test('setTabs', function(assert) {
+        var config = {
+        };
```
ferenckiss89: remove enter

src/tabs.js
```diff
@@ -0,0 +1,209 @@
...
+import tabsTpl from 'ui/tabs/tpl/tabs';
+import 'ui/tabs/css/tabs.css';
+
+const ns = 'tabs';
+
+/**
+ * Default config
+ * @param {Integer} activeTabIndex - the index of the tab to start on
+ * @param {Boolean} showHideTargets - if true, no need to wire up tabs to panels, it's automatic
+ * @param {String} targetDataAttr - the data attribute suffix used to connect tabs to panels
```
ferenckiss89: I'd suggest calling this by default tab-content, panel feels a bit out of the context.

src/tabs.js
```diff
@@ -0,0 +1,209 @@
...
+     * @param {Boolean} [callOnClick=true] - if false, skips the onClick call
+     * @returns {tabs} instance
+     * @fires activate-tab
+     * @throws {TypeError} on invalid index param
+     */
+    activateTabByIndex(index, callOnClick = true) {
+        const $tabBar = this.getElement();
+        const attr = this.config.targetDataAttr;
+
+        if (!_.isNumber(index) || index < 0 || index >= tabs.length) {
```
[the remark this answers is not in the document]
oatymart: Good idea, all changed.

src/tabs.js
oatymart: You mean if it throws? I don't think that should be documented as undefined return type.

src/tabs.js
oatymart: Done. It's now `data-tab-name` on the button and `data-tab-content` on the content element. Non-configurable.

src/tabs.js
```diff
@@ -0,0 +1,201 @@
...
+ */
+/**
+ * @author Martin Nicholson <martin@taotesting.com>
+ */
+/**
+ * @example
+ * var instance = tabs({
+ *   renderTo: $container,
+ *   tabs: [
+ *     { label: 'TAO Local', name: 'local-delivery' },
```
ferenckiss89: content or contentSelector (content-selector) instead of name...label and name are kinda similar

src/tabs/scss/tabs.scss
```diff
@@ -0,0 +1,50 @@
...
+    display: inline-flex;
+    flex-direction: row;
+    flex-wrap: nowrap;
+
+    .tab {
+        button {
+            @include border-radius(0);
+            padding: 0 15px;
+            background-color: white;
+            color: $textColor !important; // override button text color from basic.scss
```
ferenckiss89: Maybe important is not needed (would be better to avoid it)

src/tabs.js
```diff
@@ -0,0 +1,201 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2019 (original work) Open Assessment Technologies SA ;
+ */
+/**
+ * @author Martin Nicholson <martin@taotesting.com>
+ */
+/**
+ * @example
+ * var instance = tabs({
```
ferenckiss89: const

src/tabs/scss/tabs.scss
oatymart: You are right.

src/tabs.js
oatymart: Discussed, but we settled on using `name` for now.

PR #47, opened 2019-07-26
add update method to taskcreationbutton

ferenckiss89:
Jira: https://oat-sa.atlassian.net/browse/NEX-244

- updateTaskCreationUrl has been added to the selector component to be able to update the url dynamically

Post-merge:
- [ ] release on github
- [ ] publish to npm registry

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "0.20.0",
+    "version": "0.20.1",
     "displayName": "TAO Core UI",
     "description": "UI libraries of TAO",
     "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "0.20.1",
+    "version": "0.21.0",
     "displayName": "TAO Core UI",
     "description": "UI libraries of TAO",
     "scripts": {
--- a/src/destination/selector.js
+++ b/src/destination/selector.js
@@ -76,8 +76,9 @@ export default function destinationSelectorFactory($container, config) {
              * @param {String} url - url of the task creation
              */
             updateTaskCreationUrl: function updateTaskCreationUrl(url) {
-                if (this.taskCreationButton && this.taskCreationButton.config) {
+                if (this.config && this.taskCreationButton && this.taskCreationButton.config) {
                     this.taskCreationButton.config.taskCreationUrl = url;
+                    this.config.taskCreationUrl = url;
                 }
             }
         },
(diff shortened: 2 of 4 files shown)


src/destination/selector.js
```diff
@@ -165,14 +174,14 @@ export default function destinationSelectorFactory($container, config) {
             //enable disable the action button
             this.resourceSelector.on('change', function(selected) {
                 if (selected && _.size(selected) > 0) {
-                    button.enable();
+                    self.taskCreationButton.enable();
 
                     //append the selected class URI to the task creation data
                     if (_.isPlainObject(button.config.taskCreationData)) {
-                        button.config.taskCreationData.classUri = getSelectedUri();
+                        self.taskCreationButton.config.taskCreationData.classUri = getSelectedUri();
```
oatymart: In the condition, there's an instance of `button` you forgot to change.

src/destination/selector.js
```diff
@@ -69,6 +69,16 @@ export default function destinationSelectorFactory($container, config) {
...
                 }
+            },
+
+            /**
+             * Updates the url on taskCreationButton
+             * @param {String} url - url of the task creation
+             */
+            updateTaskCreationUrl: function updateTaskCreationUrl(url) {
+                if (this.taskCreationButton && this.taskCreationButton.config) {
+                    this.taskCreationButton.config.taskCreationUrl = url;
```
oatymart: I think we should also update `this.config.taskCreationUrl` with the new value, to avoid being out of sync.

test/destination/selector/test.js
```diff
@@ -54,7 +54,9 @@ define(['jquery', 'ui/destination/selector', 'json!test/ui/destination/selector/
             { title: 'getContainer' },
             { title: 'getElement' },
             { title: 'getTemplate' },
-            { title: 'setTemplate' }
+            { title: 'setTemplate' },
+            { title: 'update' },
+            { title: 'updateTaskCreationUrl' }
```
oatymart: Could we unit test the new method?

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "0.20.0",
+    "version": "0.20.1",
```
oatymart: Will need another bump when you're done, I think to `0.21.0` since this can be considered a feature.

src/destination/selector.js
ferenckiss89: true, fixing

src/destination/selector.js
ferenckiss89: good idea (however ultimately it is not really used)

package.json
ferenckiss89: yes

test/destination/selector/test.js
ferenckiss89: done

PR #54, opened 2019-07-31
update lib and integrate qunit-test-runner

btamas:
Update tao-core-libs and remove local libs.
Also integrato tao-qunit-testrunner

Requirement:
- [ ] https://github.com/oat-sa/tao-core-libs-fe/pull/6
- [ ] Update tao-core-libs version in package.json
- [ ] Update package-lock.json


merge.sh
```diff
@@ -0,0 +1,2 @@
+cp -r ui_should/$1 src
```
krampstudio: I think those files shouldn't have been commited

fileMerge.sh
```diff
@@ -0,0 +1 @@
+npx amdtoes6 ui_should/$1.js > src/$1.js && rm ui_should/$1.js
```
krampstudio: this file seems to be external tooling

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM node:8-slim
```
krampstudio: I think we should rely on only one image instead of maintaining Dockerfile in every repo

btamas: These files are unfortunately committed files. Thanks to notice!

PR #59, opened 2019-08-16
Fix/tao 8725/update interactjs

handleman:
**task:** https://oat-sa.atlassian.net/browse/TAO-8725
**description:** 
on hybrid laptops ( both with keyboard and touchscreen) all plugins (widgets) of test runner were unable to move through the touchscreen instead of mouse or touchpad. 

**solution**
interactjs lib which is responsible for all interface interactions was updated to newer version which supports modern " pointer events" browser API which solved the issue.

**related PR**
https://github.com/oat-sa/tao-test-runner-qti-fe/pull/21
https://github.com/oat-sa/tao-core-libs-fe/pull/8

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -866,8 +866,9 @@
             }
         },
         "@oat-sa/tao-core-libs": {
-            "version": "github:oat-sa/tao-core-libs-fe#9bdf68b89c1bf3b3ddbeadbcfc74047e98e95ace",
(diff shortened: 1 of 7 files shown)


oatymart:
@handleman The unused import `import triggerMouseEvent from 'core/mouseEvent';` in `interactUtils.js` now causes a rollup warning when I build the code.

PR #61, opened 2019-08-20
Feature/nex 297 add setconfig to themes handler

btamas:
Add `setConfig` functionality to theme handler.

Test: `npm run test`

After merge:
- [ ] release it to npm registry

### DIFF
--- a/src/themes.js
+++ b/src/themes.js
@@ -40,13 +40,19 @@ export default {
      * @returns {Object}
      */
     getConfig: function getConfig() {
-        var initialConfig;
-
+        // set theme config from module.config() if it is not defined
         if (!themesConfig) {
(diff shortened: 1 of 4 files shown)


src/themes.js
```diff
@@ -40,15 +40,21 @@ export default {
...
+            this.setConfig(module.config());
         }
         return themesConfig;
     },
 
+    /**
+     * Set themes config
+     * @param {Object} config Themes config
+     */
+    setConfig(config) {
```
ferenckiss89: if this sets only the theme config then I'd call it setThemeConfig

btamas: Based on our discussion, I will keep the name of the function.

PR #62, opened 2019-08-20
update tooltip text

ferenckiss89:
Jira: https://oat-sa.atlassian.net/browse/TAO-6397

- update tooltip warning text for non-numeric values

Post-merge:
- [ ] release on github
- [ ] publish to npm
- [ ] create pr to update version in tao-core

### DIFF
--- a/src/validator/validators.js
+++ b/src/validator/validators.js
@@ -48,7 +48,7 @@ var _validatePattern = function _validatePattern(value, callback, options) {
 var validators = {
     numeric: {
         name: 'numeric',
-        message: __('must be numeric'),
+        message: __('The value of this field must be numeric'),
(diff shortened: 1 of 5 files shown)


src/validator/validators.js
```diff
@@ -48,7 +48,7 @@ var _validatePattern = function _validatePattern(value, callback, options) {
 var validators = {
     numeric: {
```
oatymart: It seems that there is no unit test coverage of the `numeric` validator at all. Could we take this opportunity to add one?

ferenckiss89: This was not planned in the 1sp, I extended the existing one to check for the numeric selector for the validator.

PR #65, opened 2019-10-28
Fix/nex 511 fix webpack build warnings

btamas:
Related to https://oat-sa.atlassian.net/browse/NEX-511

Remove old not supported flex css style definitions.

Related PR:
- [ ] https://github.com/oat-sa/tao-deliver-testrunner-fe/pull/42

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "0.24.1",
+    "version": "0.24.2",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "0.24.0",
+    "version": "0.24.1",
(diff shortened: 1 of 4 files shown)


btamas:
> Good job, I don't see Webpack warnings any more.
> 
> Although, when I `npm i` this package I see a rollup warning about `flatpickr.css`:
> 
> ```
> ~/work/htdocs/deliver/tao-deliver-testrunner-fe/node_modules/@oat-sa/tao-core-ui/src/datetime/picker.js → dist/datetime...
> lib/flatpickr/flatpickr.css was not found!
> created dist/datetime in 165ms
> ```
> 
> It's a strange one because the file importing it finds the other `flatpickr` imports. Can anything be done?

The reason why it is there, because this is the only css file what is loaded from node_modules. So the `css` rollup plugin should left this file there, because the project, where tao-core-ui is installed, will install `flatpickr` lib and provide css as well. Maybe I can play with the alias and if it is points to node_modules and the required file is css then I can remove the warning.

krampstudio:
> Everything looks good!
> Just one thing: I have a windows machine and `npm install` only runs without problems on my linux subsystem.

can you please elaborate, what's the issue ?

PR #70, opened 2019-11-18
Feature/uio 1 create integer validator

btamas:
Related to https://oat-sa.atlassian.net/browse/UIO-1

Implement `integer` validator. I was not able to use `Number.isInteger`, because of IE11, so I used the polyfill from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#Polyfill

Test:
`npm run test`

- [ ] Publish to npm registry

### DIFF
--- a/src/validator/validators.js
+++ b/src/validator/validators.js
@@ -78,6 +78,17 @@ var validators = {
             }
         }
     },
+    integer: {
+        name: 'integer',
+        message: __('The value of this field must be an integer'),
+        options: {},
+        validate: function(value, callback) {
+            var r = typeof value === 'number' && isFinite(value) && Math.floor(value) === value;
+            if (typeof callback === 'function') {
+                callback.call(null, r);
+            }
+        }
+    },
     notEmpty: {
         name: 'notEmpty',
         message: __('this is required'),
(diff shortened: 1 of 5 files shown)


test/validator/test.js
```diff
@@ -1,4 +1,4 @@
-define(['lodash', 'jquery', 'ui/validator'], function(_, $, FormValidator) {
```
ferenckiss89: license header is missing from this one

src/validator/validators.js
```diff
@@ -78,6 +78,18 @@ var validators = {
             }
         }
     },
+    integer: {
+        name: 'integer',
+        message: __('The value of this field must be an integer'),
+        options: {},
+        validate: function(value, callback) {
```
krampstudio: can you please use the shorthand notation ?

src/validator/validators.js
```diff
@@ -78,6 +78,18 @@ var validators = {
...
+            const parsedValue = parseInt(value, 10);
+            var r = value === parsedValue.toString() && isFinite(parsedValue) && Math.floor(parsedValue) === parsedValue;
```
krampstudio: `const result` no ?

PR #71, opened 2019-11-26
added getKeyNavigator in API of navigableGroupElement

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-9506

Added `getKeyNavigator` in API of `navigableGroupElement`


jsconan:
Forgot to mention, but the branch name was not accurate: it should have been prefixed by `feature` instead of `fix` to respect the version change

PR #79, opened 2020-01-09
fix partials

krampstudio:


build/rollup.config.js
```diff
@@ -89,6 +89,8 @@ export default inputs.map(input => {
                     module: Handlebars
                 },
                 helpers: ['build/tpl.js'],
+                isPartial : () => true,
+                partialRoot: srcDir,
```
oatymart: The new system works well with or without these 2 lines. This part of the helper library doesn't seem to do anything now that we're defining the partial names as module paths.

build/path.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
         ui: srcDir,
         build: path.resolve(rootPath, 'build'),
         // Hardcoded paths to partial Handlebars templates (prevents unresolved template dependencies)
-        'ui-generis-widget-label.tpl': path.resolve(srcDir, 'generis/widget/widget.tpl'),
-        'ui-form-widget-label.tpl': path.resolve(srcDir, 'form/widget/tpl/label.tpl')
+        //'ui-generis-widget-label.tpl': path.resolve(srcDir, 'generis/widget/widget.tpl'),
+        //'ui-form-widget-label.tpl': path.resolve(srcDir, 'form/widget/tpl/label.tpl')
```
oatymart: Ok to remove.

PR #93, opened 2020-02-07
Fix/NCC-63/provide proper calendar locale

andrey-ponamarev:
**Related to task:**  https://oat-sa.atlassian.net/browse/NCC-63
**Description:** Provide default `locale` value to the [flatpickr](https://flatpickr.js.org/localization/) instance from document `lang` attribute (**'default'** value used as a fallback)
Fix calendar missalignment by predefined input presentation minimum width.
Enhance `flatpickr` selectors and set back library styles which are overridden by TAO table styles
**Unit tests cli:** `npm run test`

### DIFF
--- a/src/datetime/picker.js
+++ b/src/datetime/picker.js
@@ -111,19 +111,18 @@ var supportedConstraints = ['minDate', 'maxDate', 'enable', 'disable'];
  * @param {String} locale
  * @returns {Boolean}
  */
-var hasTranslationsForLocale = function(locale) {
-    return _.isObject(flatpickrLocalization.default[locale]);
-};
+const hasTranslationsForLocale = locale => _.isObject(flatpickrLocalization.default[locale]);
 
 /**
  * Detects document language
  * @returns {String | undefined}
  */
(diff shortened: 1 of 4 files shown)


src/datetime/picker.js
```diff
@@ -106,14 +106,35 @@ var setups = {
  */
 var supportedConstraints = ['minDate', 'maxDate', 'enable', 'disable'];
 
+/**
+ * Checks translation existing for given locale
+ * @param {String} locale
+ * @returns {Boolean}
+ */
+var hasTranslationsForLocale = function(locale) {
```
btamas: `const`, you can use es6 here.

src/datetime/scss/picker.scss
```diff
@@ -75,6 +75,7 @@ $picker-field-height: 37px;
                     }
                 }
                 .cur-year {
+                    padding: 0 0 0 0.5ch;
```
btamas: `ch` is not working propertly in IE11. Did you check it?

PR #94, opened 2020-02-10
Fix/ACTP-224/Translate english text on spanish interface

shpran:
Related to: [https://oat-sa.atlassian.net/browse/ACTP-224](https://oat-sa.atlassian.net/browse/ACTP-224)

Use translation function for the next variables: **resourceType** and **resourceTypes**.
Translations to the related words (**session** and **sessions**) added in the **tao** extension (es-MX).

#### How to test
 
Precondition:
* Change proctor language interface to `Mexican Spanish`;
* You need to have some test-takers on proctor screen;
* Use the following command to update translations: `php tao/scripts/taoUpdate.php`.

Log in as proctor.
On the proctor screen select one or multiple test-takers and try to authorize/terminate them.
You will see a model with list of test-takers that will be processed (or not) during this action.

Requires :
 - [ ] https://github.com/oat-sa/tao-core/pull/2393

### DIFF
--- a/src/bulkActionPopup/tpl/layout.tpl
+++ b/src/bulkActionPopup/tpl/layout.tpl
@@ -5,13 +5,13 @@
         {{#if single}}
     <div class="single" data-resource="{{allowedResources.0.id}}">
         <p>
-            {{__ "The action will be applied to"}} {{resourceType}} <span class="resource-label">{{allowedResources.0.label}}</span>
+            {{__ "The action will be applied to"}} {{__ resourceType}} <span class="resource-label">{{allowedResources.0.label}}</span>
         </p>
     </div>
         {{else}}
     <div class="multiple">
         <p>
-            {{__ "The action will be applied to the following"}} <span class="resource-count">{{resourceCount}}</span> {{resourceTypes}}:
+            {{__ "The action will be applied to the following"}} <span class="resource-count">{{resourceCount}}</span> {{__ resourceTypes}}:
         </p>
         <ul class="plain applicables">
             {{#each allowedResources}}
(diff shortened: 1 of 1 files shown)


src/bulkActionPopup/tpl/layout.tpl
```diff
@@ -5,13 +5,13 @@
         {{#if single}}
     <div class="single" data-resource="{{allowedResources.0.id}}">
         <p>
-            {{__ "The action will be applied to"}} {{resourceType}} <span class="resource-label">{{allowedResources.0.label}}</span>
+            {{__ "The action will be applied to"}} {{__ resourceType}} <span class="resource-label">{{allowedResources.0.label}}</span>
```
zagovorichev: I have never seen that we translate variables, I prefer to have condition here, and translate strings

zagovorichev: or translation should be passed in the variable from the script

zagovorichev: You have to close this PR, and delete branch, 
then you can find all places, like:
https://github.com/oat-sa/extension-tao-proctoring/blob/454869ba18a1d6e00db0b655b99a8a1e74e8dfa5/views/js/controller/Delivery/index.js#L255
and make there translations
as it was done for:
https://github.com/oat-sa/extension-tao-proctoring/blob/454869ba18a1d6e00db0b655b99a8a1e74e8dfa5/views/js/controller/Delivery/index.js#L253

zagovorichev: this search results can help you :
https://github.com/search?q=org%3Aoat-sa+bulkActionPopup&type=Code

shpran: I think that it is bad idea to translate `resourceType: ‘session’`.

Translation of `session` will be `sesión`
Translation of `sessions` will be `sesiones`

If you look to the file [bulkActionPopup.js](https://github.com/oat-sa/tao-core-ui-fe/blob/4d765561878d009eca17c4557005e6bb84dd6b9a/src/bulkActionPopup.js#L272) you will see, that variable `resourceTypes` is a concatenation of variable `config.resourceType` and letter `s`.
When we will translate `resourceType: __(‘session’)` there will be wrong translation - we will receive word `sesións` (not `sesiones` as we need).


And I really don’t understand why we can’t translate variable inside tpl file?
I was check it and everything was fine.
For example, in PHP we can pass function argument as variable or as plain text:
```
function bar(string $variable): void {}

$test = ‘Some text’;
bar($test);

bar(‘Plain text’);
```
I think here is the similar situation.

zagovorichev: I'm as a developer expect to see the value of what I sent, not the translation of this

shpran:
Not relevant.
Translation will be added in [another PR](https://github.com/oat-sa/extension-tao-proctoring/pull/767)

PR #114, opened 2020-04-01
Fix/actp 373/hit enter show password

AnastasiaArcadia:
**Related to:**
https://oat-sa.atlassian.net/browse/ACTP-373

**Description:**
Login page: KB focus + hit "Enter" on "Show password" button triggers login

**How to check:**
Open TAO login.
Fill in some credentials.
Set focus to Show/Hide password button using KB.
Hit Enter button.
Password should be revealed, focus on password field.

### DIFF
--- a/src/login/login.js
+++ b/src/login/login.js
@@ -199,7 +199,7 @@ export default function loginFactory($container, config) {
 
             hide();
 
-            $form.on('keypress', function (e) {
+            $inputToggle.on('keypress', function(e) {
                 var $active = document.activeElement;
                 if (e.key === 'Enter') {
                     if ($active === $hideIcon[0]) {
@@ -213,7 +213,7 @@ export default function loginFactory($container, config) {
                 }
             });
 
-            $inputToggle.on('click', function () {
+            $inputToggle.on('click', function() {
                 if ($pwdInput.type === 'password') {
                     show();
                 } else {
@@ -221,7 +221,7 @@ export default function loginFactory($container, config) {
                 }
             });
 
-            $inputToggle.on('keyup', function (e) {
+            $inputToggle.on('keyup', function(e) {
                 if (e.key === ' ') {
                     if ($pwdInput.type === 'password') {
                         show();
@@ -242,7 +242,7 @@ export default function loginFactory($container, config) {
                 messages.error = __('All fields are required');
             }
 
-            _.forEach(messages, function (message, level) {
+            _.forEach(messages, function(message, level) {
                 if (message) {
                     feedback()
                         .message(level, message)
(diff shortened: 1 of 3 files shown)


src/login/login.js
```diff
@@ -199,6 +199,20 @@ export default function loginFactory($container, config) {
 
             hide();
 
+            $inputToggle.on('keypress', function(e) {
```
jsconan, line 202: A similar code exists to apply on the spacebar ([see line 224](https://github.com/oat-sa/tao-core-ui-fe/blob/3795ab7a19a285de046c2d17dd7edf718003ccaa/src/login/login.js#L224)).

A better approach would have bee to factorize the behavior:
```javascript
function togglePassword() {
    if ($pwdInput.type === 'password') {
        show();
    } else {
        hide(true);
    }
}

$inputToggle.on('keyup', function (e) {
    if (e.key === ' ') {
        togglePassword();
    }
});
$inputToggle.on('keypress', function (e) {
    if (e.key === 'Enter') {
        togglePassword();
        e.stopPropagation();
    }
});
```

src/login/login.js
```diff
@@ -190,6 +190,14 @@ export default function loginFactory($container, config) {
                 }
             };
 
+            var togglePassword = function togglePassword() {
```
ampaveliev, line 193: const

src/login/login.js
```diff
@@ -199,23 +207,22 @@ export default function loginFactory($container, config) {
...
-                        hide(true);
-                    }
+            $inputToggle.on('keypress', function (e) {
+                if (e.key === 'Enter') {
+                    togglePassword();
+                    e.stopPropagation();
                 }
             });
+
+            $inputToggle.on('click', function() {
```
ampaveliev, line 223: No needs to create anonymous  wrapper function

src/login/login.js
```diff
@@ -199,23 +207,22 @@ export default function loginFactory($container, config) {
...
             });
 
-            $inputToggle.on('keyup', function(e) {
-                if (e.key === ' ') {
-                    if ($pwdInput.type === 'password') {
-                        show();
-                    } else {
-                        hide(true);
-                    }
+            $inputToggle.on('keypress', function (e) {
```
ampaveliev, line 216: Can  `keyup` event be used instead?

ampaveliev:
Also I have two broken tests but the errors persist in the develop branch

src/login/login.js
AnastasiaArcadia: keypress event is used in form submitting, in case of using keup submit will be done before
https://github.com/oat-sa/tao-core-ui-fe/blob/0b86959071db67fd6f9acb41538b26e225cd5091/src/login/login.js#L290

PR #116, opened 2020-04-06
Feature/actp 308/dialog fixes

ampaveliev:
**Jira ticket**
https://oat-sa.atlassian.net/browse/ACTP-308

**Description**
The screen reader user does not know what kind of element the notification is.

When a notification appears, the focus is automatically shifted to the OK button. This means that the screen reader will not actually read out the warning text within the notification but only “OK, button”. This is not alarming and very surprising to the screen reader user.

The close button does not have a focus style. The keyboard user cannot see which element has focus and what happens after hitting ENTER or SPACE.

**How to test**
1. Update `config/taoQtiTest/testRunner.conf.php`. So the value sjould be
```php
            'dialog' => array(
                'alert' => array(
                    'focus' => 'navigable-modal-body'
                ),
                'confirm' => array(
                    'focus' => 'navigable-modal-body'
                )
            ),
```
2. Create a test
3. Start the test as test taker
4. Exit for the test . You will see a dialog box

**How to run unit tests**
`npm test`


src/dialog/tpl/body.tpl
```diff
@@ -1,4 +1,8 @@
-<div class="preview-modal-feedback modal {{class}}">
+<div
+    class="preview-modal-feedback modal {{class}}"
+    role="dialog"
+    class="navigable-modal-body"
```
jsconan, line 4: duplication of the `class` attribute

PR #117, opened 2020-04-06
Breaking/actp 429/keynavigation issue in native mode

jsconan:
Related to: https://oat-sa.atlassian.net/browse/ACTP-429

Related to:
- [ ] https://github.com/oat-sa/tao-test-runner-qti-fe/pull/159

Rework the `keyNavigation` architecture.

Breaking changes:
- the `navigableDomGroup` module has been removed, the `keyNavigator` being now able to be managed as a navigable element as well no need to have an extraneous object to do so.
- the `getNavigables()` method has been removed to `getNavigableElements()`. Consumers have been updated (`dialog`, others will be in the companion PR).
- the instances of the `keyNavigator` are not persisted anymore inside a module variable
- the `replace` config property has been removed since the module variable has been removed
- the `goto()` API has been removed. It was relying on the bad design lead by the module variable. However, it was not used.
- the `get()` API has been removed, but wasn't used. Still the same reason: removal of the module variable.

The `keyNavigator` and the `navigableDomElement` are now sharing the same API, so that each of them can be managed as a navigable element.

The keyboard management has been moved from the `keyNavigator` to the `navigableDomElement` in order to improve consistency and atomicity as well as the separation of concerns. Now each module only manages its own features.

The `navigableDomElement` represents a focusable element, and it triggers events on `focus`, `blur` and `key`.

The `keyNavigator` represents a group of focusable elements. It manages the life cycle and allows navigating between the elements. Since it also includes the API of the focusable elements, a `keyNavigator` can be composed of other instances of `keyNavigator`.

Impacted unit tests:
- `test/keyNavigation/keyNavigator/test.html`
- `test/keyNavigation/navigableDomElement/test.html`
- `test/dialog/test.html`

To test:
```
rm -rf dist
npm ci
npm run test
```


ampaveliev:
Also I see on my local env an error in the browser console `core/eventifier navigable.init is not a function TypeError: navigable.init is not a function`. It rises at the next step after Next Section button has been initialized. Code line: [keyNavigation/navigator.js#L180](https://github.com/oat-sa/tao-core-ui-fe/blob/master/src/keyNavigation/navigator.js#L180 ), the `navigable.getId()` returns 
`"header-toolbar-0"`.

PR #131, opened 2020-05-25
Feature/tca 595/improve bundle and package scripting

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TCA-595

Not directly related to the ticket, but this PR contains some improvement to ease the work:
- add missing NPM scripts, to align with other packages (like tao-test-runner-qti).
    - `npm run test:dev`
    - `npm run build:dev`
- clear the output folder before building, to avoid ghost content


build/rollup.config.js
```diff
@@ -29,6 +30,8 @@ import wildcardExternal from '@oat-sa/rollup-plugin-wildcard-external';
 const { srcDir, outputDir, aliases } = require('./path');
 const Handlebars = require('handlebars');
 
+const isDev = process.env.NODE_ENV === 'development';
```
btamas, line 33: It is working, but I think it is the opposite what we do normally.
If you build something, you build it with development mode, expect you provide `NODE_ENV=production`.

jsconan: Yes, I agree. The idea here is to offer the same scripts as we have in other repos.
This code was taken from what has been done in [`tao-test-runner-qti`](https://github.com/oat-sa/tao-test-runner-qti-fe/blob/master/build/rollup.config.js#L33).
I think the reason why it has been done this way is that the default action is to build in production mode, then it may be called by other actions (like `build:dev`). 

Anyway, we need to have consistency in the scripts offered by our npm repos. This option is not mandatory to offer at least the same commands, so that I can remove this env switch for now. And we can foresee a TDD ticket to align properly all scripts across the repos.

btamas: I think because of sensitive information, like it can contain local directory structure. I have no other reason agains it.

PR #132, opened 2020-05-26
Fix/tca 418/close button login error feedback not respond kb

ampaveliev:
**Related**
https://oat-sa.atlassian.net/browse/TCA-418

**Description**
Close button in Login error feedback does not respond to KB actions

**How to test**

1. open login page
2. enter invalid credentials
3. hit Login
4. navigate to the 'x' button in the error message popup
5. hit Spacebar / Enter on the 'x' button


src/feedback.js
```diff
@@ -259,10 +261,18 @@ var feedbackFactory = function feedbackFactory($container, config) {
...
-            });
+            $closer
+                .off('click')
+                .on('click', function(e) {
+                    e.preventDefault();
+                    self.destroy();
+                })
+                .off('keyup')
+                .on('keyup', function(e) {
+                    if ([codeEnter, codeSpace].includes(e.which)) {
```
jsconan, line 272: Caution, `includes` is not supported by IE11. Please check it is properly polyfilled by the bundling process. Otherwise, you should use something else.

Aso, `event.which` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which) and should not be used anymore. Prefer `event.key` instead.

src/feedback.js
```diff
@@ -259,10 +261,18 @@ var feedbackFactory = function feedbackFactory($container, config) {
...
 
-            $closer.off('click').on('click', function(e) {
-                e.preventDefault();
-                self.destroy();
-            });
+            $closer
+                .off('click')
+                .on('click', function(e) {
+                    e.preventDefault();
+                    self.destroy();
```
jsconan, line 268: why not use arrow function? Then you'll be able to get rid of the `self` variable here.

src/feedback.js
```diff
@@ -259,10 +261,18 @@ var feedbackFactory = function feedbackFactory($container, config) {
...
+                })
+                .off('keyup')
+                .on('keyup', function(e) {
```
jsconan, line 271: arrow function should be preferred here

src/feedback.js
ampaveliev: From [jQuery documentation](https://api.jquery.com/keyup/)
> To determine which key was pressed, examine the event object that is passed to the handler function. While browsers use differing properties to store this information, jQuery normalizes the .which property so you can reliably use it to retrieve the key code. This code corresponds to a key on the keyboard, including codes for special keys such as arrows.

Hope, it'd be fixed there. Anyway it'a good point to use `event.key` and `event.code`. But it doesn't use ascii codes for enter and space keys, so I don't wan't to focus on it so far

PR #141, opened 2020-06-16
Fix TAO 10122 - IE11 graphic gap interaction when navigate to question since response

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/TAO-10122

**Description**

InternetExplorer: Have created a Graphic Gap interactions with images and matched them in response
When I do different actions with functionality, at some point, I can't seem to edit the size of the "Gap match slots" 
I am not experiencing the same problems in f.eks Chrome

Workaround: Click outside the interaction and try again

**How to test it**

- Add graphic 'Gap match' interaction to the canvas in the Internet Explorer browser
- Select a file and click 'Select'
- Click the "Plus in circle" icon
- Select `file` and click 'Select' 
- Click 'Response' and set the added picture to the gap 
-  Click 'Question' and find media sizer in the Interaction properties panel

**Experienced Behavior**

The interaction properties panel is missing.

**Expected Behavior**

The interaction properties panel is displayed as soon as 'Question' is clicked in interaction.

### DIFF
--- a/src/interactUtils.js
+++ b/src/interactUtils.js
@@ -71,6 +71,8 @@ interactHelper = {
      */
     tapOn: function tapOn(element, cb, delay) {
         var domElement,
+            firstEvent,
+            secondEvent,
             eventOptions = {
                 bubbles: true,
                 pointerId: 1,
@@ -83,8 +85,6 @@ interactHelper = {
         if (element) {
             domElement = element instanceof $ ? element.get(0) : element;
 
-            var firstEvent;
-            var secondEvent;
             if (navigator.userAgent.indexOf('MSIE') !== -1 || navigator.appVersion.indexOf('Trident/') > 0) {
                 firstEvent = document.createEvent('HTMLEvents');
                 firstEvent.initEvent('pointerdown', false, true);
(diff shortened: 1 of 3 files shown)


src/interactUtils.js
```diff
@@ -24,6 +24,7 @@ import $ from 'jquery';
 import _ from 'lodash';
 import interact from 'interact';
 import 'core/mouseEvent';
+import triggerCustomEvent from '@oat-sa/tao-core-sdk/src/core/customEvent';
```
wazelin, line 27: Shouldn't the SDK dependency be specified as a production dependency in `package.json`?

habemuscode: No longer needed. Thanks for comment

src/interactUtils.js
```diff
@@ -74,18 +74,28 @@ interactHelper = {
...
                 height: 100,
                 isPrimary: true
             };
         if (element) {
             domElement = element instanceof $ ? element.get(0) : element;
 
-            domElement.dispatchEvent(new PointerEvent('pointerdown', eventOptions));
-            domElement.dispatchEvent(new PointerEvent('pointerup', eventOptions));
+            var firstEvent;
+            var secondEvent;
```
wazelin, line 87: `var` variable declarations SHOULD be on the top of the scope.

Please either transform the whole thing into ES6, or move them upper.

habemuscode: Done

PR #148, opened 2020-06-24
[TCA-634] Support Browser Zooming Customizations

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/TCA-634
 
once the tool is enabled make sure that it fits container boundaries
 
#### How to test
 
- prepare an instance of TAO
- prepare a delivery with enabled calculator and scratchpad
- execute delivery as a test taker
- zoom delivery up to 200%
- enable calculator and scratchpad make sure that it fits container boundaries
   
Companion PR :
 - [ ] https://github.com/oat-sa/tao-core/pull/2570

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.5.0",
+    "version": "1.5.1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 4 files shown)


atsymuk:
> The code looks ok.
> However, the fix does not always work properly.
> 
> There is an issue with the resizing, the content may not follow the frame. For instance with the calculator:
> image

it seems to be a calculator specific case. there is some logic applied on `resize` event.
should be fixed by https://github.com/oat-sa/tao-core-ui-fe/pull/148/commits/9d796b49195cb85a7e28cfe5cd4a9b41d964f130

PR #153, opened 2020-07-09
UNO 401 - Proper search text box validation

habemuscode:
**FE fix for this PR**

https://github.com/oat-sa/tao-core/pull/2591

**Related to**

[UNO-401](https://oat-sa.atlassian.net/browse/UNO-401)

**Description**

Now search text box extends `MultipleElement` class and overrides `getRawData`, that returns list to the validator instead of one value. Also, PSR improvements applied.

**Steps to reproduce**

- Create a new Item or a test property via Manage Schema on a directory
- Set the property's type to List – Multiple choice – Search input
- Set the property's list relation
- Set the property Validation Rules to `notEmpty`
- Add a new item/test
- Add a few values to the input
- Try saving the item/test

**Expected**

The item/test is saved without any errors

**Experienced**

This field is required validation error appears next to the field

**Suggested solution**

class.Searchtextbox.php should inherit the class.MultipleElement.php and implement the following method.

```
public function getRawValue()
{
    return $this->values;
}
```

### DIFF
--- a/scss/inc/_forms.scss
+++ b/scss/inc/_forms.scss
@@ -342,7 +342,7 @@ input[type="password"].viewable-hiddenbox-input {
         }
     }
 
-    + .form-error {
+    & + .form-error {
         color: $error;
         text-align: right;
     }
--- a/scss/inc/_forms.scss
+++ b/scss/inc/_forms.scss
@@ -342,7 +342,13 @@ input[type="password"].viewable-hiddenbox-input {
         }
     }
 
-    & + .form-error {
+    &.error {
+        .select2-choices {
+            border-color: $error;
+        }
+    }
+
+    + .form-error {
         color: $error;
         text-align: right;
     }
--- a/scss/inc/_forms.scss
+++ b/scss/inc/_forms.scss
@@ -343,8 +343,10 @@ input[type="password"].viewable-hiddenbox-input {
     }
 
     &.error {
-        .select2-choices {
-            border-color: $error;
+        .select2-container {
+            > :first-child {
+                border-color: $error;
+            }
         }
     }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.5.3",
+    "version": "1.5.4",
     "lockfileVersion": 1,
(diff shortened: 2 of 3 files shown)


scss/inc/_forms.scss
```diff
@@ -342,7 +342,7 @@ input[type="password"].viewable-hiddenbox-input {
         }
     }
 
-    + .form-error {
+    & + .form-error {
         color: $error;
         text-align: right;
     }
```
Alroniks, line 348: I think we can add such rules for cover select2 elements
```
& + .select2-choices {
  border-color: $error;
  + .select2-input {
       color: $error;
   }
}
```

scss/inc/_forms.scss
```diff
@@ -342,7 +342,7 @@ input[type="password"].viewable-hiddenbox-input {
         }
     }
 
-    + .form-error {
+    & + .form-error {
```
wazelin, line 345: This does nothing. The compiled CSS will be identical for the sibling selector.

scss/inc/_forms.scss
habemuscode: In order to avoid to many nested rules and still support `+ .form-error {` that it works, I push another solution :)

scss/inc/_forms.scss
habemuscode: I pushed the final solution with less nested rules and still support old code that it works

scss/inc/_forms.scss
```diff
@@ -342,6 +342,12 @@ input[type="password"].viewable-hiddenbox-input {
         }
     }
 
+    &.error {
+        .select2-choices {
+            border-color: $error;
+        }
```
wazelin, line 348:
```suggestion
        .select2-container {
            > :first-child {
                border-color: $error;
            }
        }
```

This should cover most of the `select2` rendering options. It may have not only the multiple, but a single choice in the future.

habemuscode:
Double-checked, it works. Thanks @wazelin

wazelin:
> Double-checked, it works. Thanks @wazelin

I know it does, otherwise I wouldn't have suggested it ;)

Alroniks:
@habemuscode Let's go to merge and release 👍

PR #156, opened 2020-07-22
Fix BSA 183 - Allow ckeditor dynamic translations

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/BSA-183

**Description**

This PR is to allow dynamic translations of CKEditor based on selected language by the user

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.5.4",
+    "version": "1.6.0",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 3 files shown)


bziondik:
@habemuscode When I tested locally I found that FR translation didn't apply in item Authoring page on 4 buttons (`Insert Math Expression`, `Insert Shared Stimulus `...) 
I suppose these messages not from CKeditor, but from tao-core
image
In item Preview of Extended Text interaction, 2 buttons not translated `Insert Tab` `Remove Tab`
Maybe these messages also not from CKeditor
image

habemuscode:
@bziondik For that translations will be cover here: https://github.com/oat-sa/ckeditor-dev/pull/18

PR #158, opened 2020-08-05
Feature/NCC-240/atomic data table update

andrey-ponamarev:
**Related to task:**  https://oat-sa.atlassian.net/browse/NCC-240
- https://github.com/oat-sa/extension-tao-proctoring/pull/852 - usage
- https://github.com/oat-sa/tao-core-ui-fe/pull/159 - code formatting

**Description:** To be able to see data changes in real time and don't be blocked to interact with a table actions we introduced a new option `atomicUpdate`. 
By setting this option value to `true`, the table is going to save and compare incoming data with a current data in table. In case of the number and order of rows are the same (implemented in `canApplyShallowUpdate`), the `shallowUpdate` method is trying to apply atomic update for table cells.
The `shallowUpdate` responsibility is generating and replace table cells according to the layout template.
Unfortunately we can't use the current handlebars version for generating table cell from separate template, that's why we have to repeat the generating cell logic in JS.
On the picture below these logic blocks are highlighting:
Screenshot 2020-08-05 at 11 34 28

All code of `datatable` component is covered by unit tests:
 After | Before
------------ | -------------
Screenshot 2020-08-05 at 10 39 03|Screenshot 2020-08-05 at 12 50 55

**Requires:**  The code formatting and migration to the ES6
**Unit tests cli:** npm run test


src/datatable.js
```diff
@@ -714,6 +738,116 @@ var dataTable = {
...
+                            const title = getPropertyValue('title', action, nextData);
+                            const disabled = getPropertyValue('disabled', action, nextData);
+                            const icon = getPropertyValue('icon', action, nextData);
+                            const label = getPropertyValue('label', action, nextData);
+
+                            if (!hidden) {
+                                const $actionButton = $('<button>', {
+                                    class: `btn-info small ${actionId}`,
+                                    html: `${icon ? `<span class="icon-${icon}"></span>` : ''}${label || ''}`
+                                });
```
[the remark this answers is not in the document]
andrey-ponamarev: This exactly what I wanted to do for a table cell, only one thing stopped me is possibility to reuse template in template.
Now I see that it was a good idea to create template for table cell even if we can't reuse it in `layout.tpl` 🤦‍♂️ 
Anyway, I found out that template for button is already exist, so I will use it and I'm going to leave a comment how to update this part when we will update `handlebars`

src/datatable.js
```diff
@@ -249,7 +297,7 @@ var dataTable = {
             })
             .fail(function(response, option, err) {
                 var requestErr = httpErrorParser.parse(response, option, err);
-                logger.error(requestErr.message);
+                // logger.error(requestErr.message);
```
jsconan, line 300: Is it wanted or is it a forgotten debug action?
Better to avoid having dead code, so if this was wanted, please remove the line. Otherwise please restore it :)

andrey-ponamarev: Uncomment

PR #163, opened 2020-08-18
Fix TAO 10464 - Uncaught type error upon dragging the Media Interaction on canvas

habemuscode:
**Related to this issue**

https://oat-sa.atlassian.net/browse/TAO-10464

**Description**

When the user drags the Media Interaction on the canvas, there's no file selection window opens. Instead 'Uncaught type error' appears on the browser console and the widow opens only after clicking empty space outside the interaction and clicking it again.

**Steps to reproduce**

- Go to the Items tab.
- Create a new item and go to Authoring.
- Drag and drop Media Interaction on the canvas.

**Actual result**

Error in console, the file selection window is not opened.

**Expected result**

The file selection window shows up without any errors.

### DIFF
--- a/src/mediasizer.js
+++ b/src/mediasizer.js
@@ -553,7 +553,7 @@ MediaSizer = {
 
             // compatibility layer naturalWidth|Height vs. naturalHeight
             // internally width/height are used
-            options.width = options.width || options.naturalWidth || defaults.size;
+            options.width = options.width || options.naturalWidth || options.maxWidth || defaults.size;
             options.height = options.height || options.naturalHeight || defaults.size;
             options.naturalWidth = options.naturalWidth || options.width || defaults.size;
             options.naturalHeight = options.naturalHeight || options.height || defaults.size;
--- a/src/mediasizer.js
+++ b/src/mediasizer.js
@@ -53,7 +53,8 @@ var defaults = {
     responsive: true,
     showResponsiveToggle: true,
     showReset: true,
-    showSync: true
+    showSync: true,
+    size: 100
 };
 
 /**
@@ -552,10 +553,10 @@ MediaSizer = {
 
             // compatibility layer naturalWidth|Height vs. naturalHeight
             // internally width/height are used
-            options.width = options.width || options.naturalWidth || 100;
-            options.height = options.height || options.naturalHeight || 100;
-            options.naturalWidth = options.naturalWidth || options.width || 100;
-            options.naturalHeight = options.naturalHeight || options.height || 100;
+            options.width = options.width || options.naturalWidth || defaults.size;
+            options.height = options.height || options.naturalHeight || defaults.size;
+            options.naturalWidth = options.naturalWidth || options.width || defaults.size;
+            options.naturalHeight = options.naturalHeight || options.height || defaults.size;
 
             options.hasSize =
                 options.width && options.height && _.isFinite(+options.width) && _.isFinite(+options.height);
--- a/src/mediasizer.js
+++ b/src/mediasizer.js
@@ -558,8 +558,7 @@ MediaSizer = {
             options.naturalWidth = options.naturalWidth || options.width || defaults.size;
             options.naturalHeight = options.naturalHeight || options.height || defaults.size;
 
-            options.hasSize =
-                options.width && options.height && _.isFinite(+options.width) && _.isFinite(+options.height);
+            options.hasSize = _.isFinite(+options.width) && _.isFinite(+options.height);
 
             // incomplete or conflicting configurations
             // no target provided, also no width and/or no height
--- a/src/mediasizer.js
+++ b/src/mediasizer.js
@@ -552,10 +552,10 @@ MediaSizer = {
 
             // compatibility layer naturalWidth|Height vs. naturalHeight
             // internally width/height are used
-            options.width = options.width || options.naturalWidth;
-            options.height = options.height || options.naturalHeight;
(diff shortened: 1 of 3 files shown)


src/mediasizer.js
```diff
@@ -552,10 +552,10 @@ MediaSizer = {
 
             // compatibility layer naturalWidth|Height vs. naturalHeight
             // internally width/height are used
-            options.width = options.width || options.naturalWidth;
-            options.height = options.height || options.naturalHeight;
-            options.naturalWidth = options.naturalWidth || options.width;
-            options.naturalHeight = options.naturalHeight || options.height;
+            options.width = options.width || options.naturalWidth || 100;
+            options.height = options.height || options.naturalHeight || 100;
```
wazelin, line 556: Not sure about that modification. 

It may have a bigger impact, check out the `options.hasSize` flag, which is always `true` now.

habemuscode: It's only called on `init`, the impact is limited. Also, this set a default value if in any part of TAO we don't set it properly (For me, any image, video, media needs to have default values for avoid issues like no render or don't see nothing)

I will try to found other options

habemuscode: Inside `mediaEditorComponent` we have defaults too.

```
var defaultConfig = {
    mediaDimension: {
        active: false
    }
}
```

jsconan: I'm not against the default value. However, could you use a constant instead of a magic value? Or better, rely on the `defaults` object: https://github.com/oat-sa/tao-core-ui-fe/pull/163/files#diff-411e90fba10497f847418435b5355f07R546

habemuscode:
```
var defaults = {
    disableClass: 'disabled',
    applyToMedium: true,
    denyCustomRatio: true,
    responsive: true,
    showResponsiveToggle: true,
    showReset: true,
    showSync: true,
    size: 100
};

options.width = options.width || options.naturalWidth || defaults.size;
options.height = options.height || options.naturalHeight || defaults.size;
options.naturalWidth = options.naturalWidth || options.width || defaults.size;
options.naturalHeight = options.naturalHeight || options.height || defaults.size;
```
@jsconan Like this right?

wazelin: I'm not talking about not having a default value in general, I'm talking only about this particular case.
It's obvious an undefined value of `width` and `height` was expected, at least in some cases.

Now, let's find out if they are really needed, or if we should fix the part that cannot work with the undefined value instead.

habemuscode: @wazelin I'm trying to find why we pass undefined, but looking for the future, we can fix this but the bug can appear again. If we set here, a default value, doesn't matter this case or others in the future, will not appear again.

habemuscode: @wazelin @andrey-ponamarev Happens here

imagen

`sizechange.mediasizer` is fired but interaction don't have width or height (It's on `initForm`, isn't a change like changing in properties panel)

I still suggest following my first approach for be safe of all possible bugs in future instead of fix here

jsconan: @habemuscode it is better to have a constant indeed. However, as mentioned by @wazelin, there is regression with the [hasSize property](https://github.com/oat-sa/tao-core-ui-fe/pull/163/files#diff-411e90fba10497f847418435b5355f07R561).

habemuscode: Yes I plan to refactor `options.width && options.height && _.isFinite(+options.width) && _.isFinite(+options.height);` to `_.isFinite(+options.width) && _.isFinite(+options.height);`

I will not remove all, just first 2 validations.

wazelin: OK, I don't mind having the fallback values, but having them like that makes the rendering of the interaction jump for a split second.

src/mediasizer.js
```diff
@@ -552,13 +553,12 @@ MediaSizer = {
 
             // compatibility layer naturalWidth|Height vs. naturalHeight
             // internally width/height are used
-            options.width = options.width || options.naturalWidth;
-            options.height = options.height || options.naturalHeight;
-            options.naturalWidth = options.naturalWidth || options.width;
-            options.naturalHeight = options.naturalHeight || options.height;
+            options.width = options.width || options.naturalWidth || defaults.size;
```
wazelin, line 556:
```suggestion
            options.width = options.width || options.naturalWidth || options.maxWidth || defaults.size;
```

src/mediasizer.js
habemuscode:
```
const defaultSize = 100;

        // Initialize MediaSizer
        $form.find('.media-sizer-panel').on('sizechange.mediasizer', function() {
            $(this).find('input').trigger('change');
        }).mediasizer({
            showResponsiveToggle: false,
            showSync: false,
            showReset: false,
            responsive: false,
            applyToMedium: false,
            width: interaction.object.attr('width') || defaultSize,
            height: interaction.object.attr('height') || defaultSize,
            minWidth: 50,
            maxWidth: $container.innerWidth()
        });
```

@wazelin I implement this but I don't see the difference. Can't reproduce that "jump", but OK.

I will create additional PR for that.

wazelin (approved):

src/mediasizer.js
habemuscode: https://github.com/oat-sa/extension-tao-itemqti/pull/1515

andrey-ponamarev (approved):
- [ ] The feature is working correctly on my local machine (if applicable)

PR #170, opened 2020-08-25
Feature/tao 10453/create search modal component

adriloma:
**Related to:** https://oat-sa.atlassian.net/browse/TAO-10453

**Description**
This PR creates a new `searchModal` component that extends `ui/modal`. This component will be used to manage simple and advanced search. When the search returns no results, or search has been cleaned/not triggered, information messages are displayed.

**How to test**
To test this PR locally you also need to checkout your `tao-core` repository to [implement-query-builder](https://github.com/oat-sa/tao-core/tree/feature/TA0-10453/implement-query-builder) branch. That way, when you trigger a search in tao action bar with `enter` keypress, the new component will be used. Current component can still be used clicking on search icon.

image

image

image


src/searchModal.js
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * Creates search modal
+     */
+    function initModal() {
+        instance
+        .getElement()
+        .addClass('modal')
+        .on('closed.modal', function() {
+            instance.destroy();
+        })
```
andrey-ponamarev, line 71: Please use prettier for code formatting https://github.com/oat-sa/prettier-config

src/searchModal.js
```diff
@@ -0,0 +1,207 @@
...
+                    label: 'Go to item',
+                    id: 'go-to-item',
+                    action: function openResource(id){
+                        config.events.trigger('refresh', {
+                            uri: id
+                        });
+                        destroyModal();
+                    }
+                }
+            },
```
andrey-ponamarev, line 152: Please use `actions` options as Array

```suggestion
            actions : [
                  {
                    id: 'go-to-item',
                    icon: 'open',
                    label: __('Go to item'),
                    action: function openResource(id){
                        config.events.trigger('refresh', {
                            uri: id
                        });
                        destroyModal();
                    }
                }
            ],
```

src/searchModal.js
```diff
@@ -0,0 +1,207 @@
...
+    let running = false;
+    instance.on('render', () => renderModal());
+    instance.on('destroy', () => destroyModal());
+
+    /**
+     * Creates search modal, inits template selectors, and triggers initial search
+     */
+    function renderModal() {
+        initModal();
+        initUiSelectors();
```
andrey-ponamarev, line 48: Could it be moved to the `.on('init', ` event?

src/searchModal.js
```diff
@@ -0,0 +1,207 @@
...
+ * @param {object} config.events - events hub
+ */
+export default function searchModalFactory(config) {
+    const instance = component().setTemplate(layoutTpl);
+    let searchInput = null;
+    let searchButton = null;
+    let clearButton = null;
+    let running = false;
+    instance.on('render', () => renderModal());
+    instance.on('destroy', () => destroyModal());
```
andrey-ponamarev, line 41:
```suggestion
    instance
      .on('render', renderModal);
      .on('destroy', destroyModal);
```

src/searchModal.js
```diff
@@ -0,0 +1,207 @@
...
+
+        section.empty();
+        section.append($tableContainer);
+        $tableContainer.on('load.datatable', searchResultsLoaded);
+
+        //create datatable
+        $tableContainer.datatable({
+            url: data.url,
+            model : _.values(data.model),
+            emptyText: "No results were found.",
```
andrey-ponamarev, line 137: Translation is missing

```suggestion
            emptyText: __("No results were found."),
```

src/searchModal.js
```diff
@@ -0,0 +1,207 @@
...
+            message = 'Please define your search in the search panel.';
+            icon = 'icon-find';
+
+        } else if (reason === 'no-matches') {
+            message = 'No item found. Please try other search criteria.';
+            icon = 'icon-info';
+        }
+
+        section.append(`
+        <div class='no-datatable-container'>
```
andrey-ponamarev, line 200: Could you use handlebars instead?

src/searchModal/tpl/layout.tpl
```diff
@@ -0,0 +1,16 @@
+<div class="search-modal section-container">
+    <div class="clear content-wrapper content-panel">
+        <div class="navi-container">
+            <div class="search-bar-container">
+                <span class="icon-find"></span>
+                <input type="text" placeholder="Search Item">
+            </div>
+            <div class="buttons-container">
+                <button class="btn-clear btn-transparent small">Clear</button>
+                <button class="btn-search btn-info small">Search</button>
```
andrey-ponamarev, line 10: Translations

andrey-ponamarev (changes requested):

src/searchModal.js
adriloma: I have checked, but on `init` event the component is not yet rendered, causing an error on `initModal()` when:

```
    function initModal() {
        instance
            .getElement()
            .addClass('modal')
            (...)
```
Because instance.getElement() returns `undefined`.

src/searchModal.js
adriloma: I have tried, but `prettier` keeps formating it into a single line. I have checked that the way prettier works is to compare the length of the line to the `printWidth` property of the `.prettierrc.json` and split it if the line exceeds the value. If I change it to `40`, for example, I get what you propose. So don't know how to achieve this without updating the `.prettierrc.json`. 

It seems that this behaviour was a recent [update](https://prettier.io/blog/2020/03/21/2.0.0.html) on prettier. I am ussing prettier `2.0.5`, do you know which prettier version should I use?

src/searchModal.js
adriloma: Instead I removed the property as the case for no results is managed outside the datatable, so the message will never be displayed.

src/searchModal.js
andrey-ponamarev: The suggestion is written without formatting. 
You have a several options: commit the suggestion, format the code and commit formatting or copy/paste that part and apply formatting before commit, it's up to you :)

adriloma: ops, sorry, yesterday I was blind and didn't see the actual change (to pass directly the function instead of using an anonymous function) so did not understand the change beyond splitting it in two lines. 

I commited your proposal but on a single line, as there is no method chaining there, just a simple call.

src/searchModal.js
```diff
@@ -0,0 +1,203 @@
...
+            if (running === false) {
+                running = true;
+                $.ajax({
+                    url: config.url,
+                    type: 'POST',
+                    data: { query: query },
+                    dataType: 'json'
+                })
+                    .done(buildSearchResultsDatatable)
+                    .complete(function () {
```
andrey-ponamarev, line 112: `complete()` will be deprecated, please use `always()` instead
```suggestion
                    .always(function () {
```

andrey-ponamarev (changes requested):

src/searchModal.js
```diff
@@ -0,0 +1,203 @@
...
+        initUiSelectors();
+        searchButton.trigger('click');
+    }
+
+    /**
+     * Removes search modal
+     */
+    function destroyModal() {
+        instance.getElement().removeClass('modal').modal('destroy');
+        $('.modal-bg').remove();
```
andrey-ponamarev, line 58: Should we remove `searchButton`, `clearButton` as well?

adriloma: I think it is not necessary. Both buttons, same as other html elements, are inside `.modal-bg`, so should be removed from DOM too, correct?

src/searchModal/scss/searchModal.scss
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+    }
+    .content-container {
+        .no-datatable-container {
+            display: flex;
+            flex-direction: column;
+            align-items: center;
+            .no-datatable-icon {
+                color: lightgray;
```
andrey-ponamarev, line 41: Since for different clients we're using different themes, we need to use a global css variables 🤔
Could you include on the top `@import "inc/bootstrap";` and pick the proper color name from https://github.com/oat-sa/tao-core-ui-fe/blob/master/scss/inc/_colors.scss ?

src/searchModal.js
andrey-ponamarev: Maybe 🤷‍♂️ that's why I'm asking 🙂. 
Is it possible to call the `renderModal` method twice without page reload?
In case it's possible, could you check that `search` and `clear` executes only one time per click event, otherwise it fine?
```
      searchButton.on('click', search);
      clearButton.on('click', clear);
```

src/searchModal/scss/searchModal.scss
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// Custom class for close modal button, so it poisitioned on the top-left corner instead of top-right (default)
+.modal-close-left {
+    position: absolute;
+    left: 20px;
+    top: 10px;
+    cursor: pointer;
+    font-weight: bold;
+    color: #222 !important;
```
andrey-ponamarev, line 70:
```suggestion
    color: $textColor !important;
```

src/searchModal.js
adriloma: Yep, I confirm that is only executed once even if I render the search component multiple times : )

test/search/test.js
```diff
@@ -0,0 +1,56 @@
...
+    QUnit.test('search component modal is correctly rendered', function(assert) {
+        const ready = assert.async();
+        const searchModal = searchModalFactory({
+            query: 'mathematics',
+            url: ''
+        });
+        const $container = $('.search-modal');
+        const searchInput = $container.find('.search-bar-container input');
+        const clearButton = $('.btn-clear');
+        const closeButton = $('.modal-close-left');
```
andrey-ponamarev, line 36: Weird, I thought `searchModal` should generate children components asynchronously (i.e. on render event) 🤔
It would be nice if you can add visual module and check the code coverage

andrey-ponamarev (approved):
Looks good! 
⚠️ Please bump the package version before merge.

- [ ] The feature is working correctly on my local machine (if applicable)

src/searchModal.js
```diff
@@ -0,0 +1,203 @@
...
+        } else if (reason === 'no-matches') {
+            message = __('No item found. Please try other search criteria.');
+            icon = 'icon-info';
+        }
+
+        const infoMessage = infoMessageTpl({ message, icon });
+        section.append(infoMessage);
+    }
+
+    return instance.init({ renderTo: 'body' });
```
bziondik, line 202: please, move`renderTo: 'body'` to default configs

src/searchModal.js
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Creates a search modal instance
+ * @param {object} config - search modal configuration
+ * @param {string} config.query - search query to be set and triggered on component creation
+ * @param {string} config.url - search endpoint
+ * @param {object} config.events - events hub
+ */
+export default function searchModalFactory(config) {
+    const instance = component().setTemplate(layoutTpl);
```
bziondik, line 36: maybe for next PR, please use the same structure as in other components
https://github.com/oat-sa/tao-core-ui-fe/blob/dab6658aad77dca8b7a7b5c21b07dbf8760a4318/src/bulkActionPopup.js
or
https://github.com/oat-sa/tao-core-ui-fe/blob/master/src/dropdown.js

test/search/test.js
```diff
@@ -0,0 +1,56 @@
...
+        setTimeout(function() {
+            closeButton.trigger('click');
+            setTimeout(function() {
+                const $container = $('.search-modal');
+                assert.equal($container.length, 0, 'search component modal is destroyed');
+                ready();
+            }, 500);
+        }, 500);
+    });
+});
```
bziondik, line 56: maybe for next PR, please add Visual case
https://github.com/oat-sa/tao-core-ui-fe/blob/dab6658aad77dca8b7a7b5c21b07dbf8760a4318/test/bulkActionPopup/test.js#L559

bziondik (changes requested):
I comment in Unit tests line 49 `// closeButton.trigger('click');` and style of search modal is not the same as in the design
If some styles from tao-core are applied they should be moved to `tao-core-ui-fe/src/searchModal/scss/searchModal.scss`
image

bziondik (approved):
Styles is correct
image

habemuscode:
imagen

I see some error on destroy

PR #189, opened 2020-10-01
Fix/tao 10561/improve management of private classes

adriloma:
**Related to:** https://oat-sa.atlassian.net/browse/TAO-10561

**Description**
If `taoDacSimple` extension is installed users should not be able to read classes or items with no read access. Although this was working, there was no visual indication of private classes. Now, if received class tree contains the `supportedRights` array, property `accessMode` is set to `denied` on classes with no `read` permission, so they are rendered as disabled nodes by `tree` component. 

Also item previewing was causing a style issue by setting `.navi-container` to `display:none`. This is fixed so a 
 `.navi-container` inside of a `.search-modal` always has `display: flex`.

access-control-pr

### DIFF
--- a/src/searchModal.js
+++ b/src/searchModal.js
@@ -132,6 +132,9 @@ export default function searchModalFactory(config) {
                 const route = urlUtil.route('getAll', 'RestResource', 'tao');
                 request(route, classOnlyParams)
                     .then(response => {
+                        if (response.permissions.supportedRights && response.permissions.supportedRights.length > 0) {
+                            manageClassTreePermissions(response);
+                        }
                         resourceSelector.update(response.resources, classOnlyParams);
                     })
                     .catch(e => instance.trigger('error', e));
@@ -156,6 +159,26 @@ export default function searchModalFactory(config) {
 
             setResourceSelectorUIBehaviour();
         });
+    }
+
+    /**
+     * Loops through each class in received class tree and sets access mode to 'denied' on private classes
+     * so are disabled on class tree
+     * @param {object} classTree - class tree received by server, containing resources (classes) and permissions
+     */
+    function manageClassTreePermissions(classTree) {
+        const disableBlockedClasses = function (resources) {
+            resources.forEach((resource, index, array) => {
+                if (classTree.permissions.data[resource.uri].find(permission => permission === 'READ')) {
+                    if (resource.children) {
+                        disableBlockedClasses(resource.children);
+                    }
+                } else {
+                    array[index].accessMode = 'denied';
+                }
+            });
+        };
+        disableBlockedClasses(classTree.resources);
     }
 
     /**
--- a/src/searchModal.js
+++ b/src/searchModal.js
@@ -173,7 +173,7 @@ export default function searchModalFactory(config) {
      */
     function manageClassTreePermissions(classTree) {
         const disableBlockedClasses = function (resources) {
-            resources.forEach((resource, index, array) => {
+            _.forEach(resources, (resource, index, array) => {
                 if (
                     classTree.permissions.data[resource.uri] &&
                     classTree.permissions.data[resource.uri].find(permission => permission === 'READ')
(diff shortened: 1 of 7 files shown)


src/searchModal/scss/searchModal.scss
```diff
@@ -231,7 +231,7 @@ $classTreeVariables: (
     height: 100% !important;
     padding: 0px !important;
     .navi-container {
-        display: flex;
+        display: flex !important;
```
jsconan, line 234: **nitpick:** We must avoid using the `!important` modifier. It prevents to properly extend the style. Unless in a very few edge cases, it should be avoided.
I know it is used elsewhere in TAO, but this a bad legacy.
Is there not any way to do without it?

src/searchModal.js
```diff
@@ -132,6 +132,9 @@ export default function searchModalFactory(config) {
                 const route = urlUtil.route('getAll', 'RestResource', 'tao');
                 request(route, classOnlyParams)
                     .then(response => {
+                        if (response.permissions.supportedRights && response.permissions.supportedRights.length > 0) {
```
jsconan, line 135: **though:** Is the `permission` property always set in the response?

src/searchModal.js
```diff
@@ -158,6 +161,26 @@ export default function searchModalFactory(config) {
...
     }
 
+    /**
+     * Loops through each class in received class tree and sets access mode to 'denied' on private classes
+     * so are disabled on class tree
+     * @param {object} classTree - class tree received by server, containing resources (classes) and permissions
+     */
+    function manageClassTreePermissions(classTree) {
+        const disableBlockedClasses = function (resources) {
+            resources.forEach((resource, index, array) => {
```
jsconan, line 171: **though:** Are you sure the `resources` variable is always an array?

src/searchModal.js
```diff
@@ -158,6 +161,26 @@ export default function searchModalFactory(config) {
...
+                if (classTree.permissions.data[resource.uri].find(permission => permission === 'READ')) {
```
jsconan, line 172: **though:** Are you sure the data is there? Won't it throw a runtime issue?

jsconan (commented):
Quick code review.
I didn't run the unit tests either.

src/searchModal/scss/searchModal.scss
adriloma: I changed the class name so styles are not overwritten by item preview logic

src/searchModal.js
adriloma: It seems it always set even when `taoDacSimple` is not installed, but I saw that in [other parts of TAO](https://github.com/oat-sa/tao-core/blob/c2e5436443deee674241e77c9530f184105e1328/views/js/provider/resources.js#L141) it is always checked, so added the extra condition here too.

src/searchModal.js
adriloma: It seems to always be there for this case of use, but after checking [other implementations](https://github.com/oat-sa/tao-core/blob/c2e5436443deee674241e77c9530f184105e1328/views/js/provider/resources.js#L151), now I check it first.

src/searchModal.js
adriloma: Yes, for what I have seen, [this is what is being used](https://github.com/oat-sa/tao-core/blob/master/models/classes/resources/ResourceService.php#L164) in the BE, which is called [from here](https://github.com/oat-sa/tao-core/blob/ea9abe98edbddc7d75df0c212dfb68c3b00b32e6/actions/class.RestResource.php#L147), which is the endpoint of FE request.

bziondik (approved):
Code review only
- [x]  New code is covered by tests (if applicable)
- [x]  Tests are running successfully (old and new ones) on my local machine (if applicable)
- [ ]  Feature is working correctly on my local machine (if applicable)

src/searchModal.js
jsconan: Well, ok. But the reason why of my question is there no distinction in PHP between pure array and associative array. So that a PHP array can be either translated into pure array or into an object when it comes to JSON, depending if the keys are numeric and contiguous or not.

jsconan (approved):
Quick code review only.
Looks good at a first glance.
**suggestion:** secure the code looping over resources, either by making sure it is an array, or by using a generic algo like  `_.forEarch()`

PR #203, opened 2020-10-27
fix/TR-43-adaptieve-size-for-checkboxes

meefox:
This is 1 of 4 parts that is needed for https://oat-sa.atlassian.net/browse/TR-43
It is needed to make selector independent from strip three hierarchy

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.13.0",
+    "version": "1.13.1",
     "lockfileVersion": 1,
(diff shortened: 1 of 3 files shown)


jsconan (changes requested):
Looks straightforward.
However, I see a couple of issues:
- the pull request needs to target the `develop` branch (`master` is reserved to releases).
- the pull request is lacking the version bump (`package.json` needs a version change, following the semantic versioning)

jsconan (commented):
The code looks ok, but I cannot see a clear use case for it.
I will wait for the companion PR to have a better idea of the intents.

PR #236, opened 2020-12-28
[Fix] CON-373 advanced search UI adjustments

404passionFruit:
**Issues**

https://oat-sa.atlassian.net/browse/CON-373

**Fix**

- [x] Remove the text hint for all input fields, as they are not providing any value
- [x] Improve accessibility of advanced search criterion component

**!important:** Forked from `fix/CON-122/graphic-design-match-mockup`.  https://github.com/oat-sa/tao-core-ui-fe/pull/235 needs to be merged first.

PR #240, opened 2020-12-31
[Fix] package-lock.json version

404passionFruit:
https://github.com/oat-sa/tao-core-ui-fe/pull/236/files that was recently merged is missing version update in package-lock.json

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.20.3",
+    "version": "1.20.4",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.20.3",
+    "version": "1.20.4",
```
boajer, line 3: regardless that is only FE, please bump version at `manifest` file;

bartlomiejmarszal (approved):
LGTM

package-lock.json
404passionFruit: > regardless that is only FE, please bump version at `manifest` file;

There is no `manifest` in this repo, I'll bump it in [related PR in tao-core](https://github.com/oat-sa/tao-core/pull/2769)

PR #259, opened 2021-03-19
Fix/MS-1056/highlight group index

andrey-ponamarev:
**Related to task:**  https://oat-sa.atlassian.net/browse/MS-1056
**Description:** The highlighter is applying the wrong group index if there is a space between highlights 
**How to test:** 
1. Login to MS project as a scorer
2. Use nightlight feature to highlight two parts of text with space between them
3. Reload the item task 
4. The highlights must be restored correctly

### DIFF
--- a/src/highlighter.js
+++ b/src/highlighter.js
@@ -44,7 +44,7 @@ var defaultBlackList = ['textarea', 'math', 'script', '.select2-container'];
  * @param {Object} [options.colors] - keys is keeping as the "c" value of storing/restore the highlighters for indexing, values are wrappers class names
  * @returns {Object} - the highlighter instance
  */
-export default function(options) {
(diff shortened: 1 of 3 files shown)


src/highlighter.js
```diff
@@ -682,7 +689,7 @@ export default function(options) {
      * @returns {boolean}
      */
     function isWrappable(node) {
-        return isText(node) && !isBlacklisted(node) && node.textContent.trim().length > 0;
+        return isText(node) && !isBlacklisted(node) && node.textContent.length > 0;
```
andrey-ponamarev, line 692: Here is the fix

anshul119 (approved):
Checked together with Andrii, works as expected!

PR #262, opened 2021-03-24
fix/MS-1061/Do not skip the empty text nodes during inlineRanges build

andrey-ponamarev:
**Related to task:**  https://oat-sa.atlassian.net/browse/MS-1061
**Description:** In case of two highlights has node with an empty space(could be any number of spaces or line break) between, during the building highlights indexes, the process creates inline ranges.
The inline ranges build phase were skipping the text node that empty(i.g contains space) that cause the wrong order of setting indexes.
Screenshot 2021-03-24 at 17 38 01

**How to test:** 
- Login as Scorer
- Highlight 2 or more nodes to leave an empty space between
- Reload the page.
- The highlighting parts must be restored correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.22.5",
+    "version": "1.22.6",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@oat-sa/tao-core-ui",
-            "version": "1.22.5",
+            "version": "1.22.6",
             "license": "GPL-2.0",
             "devDependencies": {
                 "@babel/core": "^7.4.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.22.5",
+    "version": "1.22.6",
     "displayName": "TAO Core UI",
     "description": "UI libraries of TAO",
     "scripts": {
--- a/src/highlighter.js
+++ b/src/highlighter.js
@@ -508,7 +508,7 @@ export default function (options) {
                 // an isolated node (= not followed by a highlight table text) with its whole content highlighted
             } else if (
                 isWrappingNode(currentNode) &&
-                !isWrappable(currentNode.nextSibling) &&
+                !isText(currentNode.nextSibling) &&
                 (!isWrappingNode(currentNode.nextSibling) ||
                     currentNode.className === currentNode.nextSibling.className)
             ) {
--- a/src/highlighter.js
+++ b/src/highlighter.js
@@ -371,7 +371,7 @@ export default function (options) {
      * @param {number} groupId - the highlight group
      */
     function wrapTextNode(node, groupId) {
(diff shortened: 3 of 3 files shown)


no-chris:
If I understand well, this relaxes the conditions under which a new set of inline ranges would be created to avoid last week's issue. So here a potential new edge case would now be that we would be _too_ relaxed and would consider two unrelated text nodes as part of the same text node? Are we confident this cannot happen?

The `isText()` check is actually already performed inside of the `isHotNode() > isWrappable()` helper, can you explain why it needs to be performed alone apart from the other checks? (especially without the `isBlacklisted()` check)? Could we create the relevant test cases maybe?

anshul119 (approved):
Code-wise it looks good. I cant test it on local because of issues with npm not being abe to install branches from github, I will check on video call with Andrii on his machine

**Review Checklist**
- [ ] New code is not subject to concurrency issues (if applicable)
- [x] Commits are following [conventional commits]

andrey-ponamarev:
> If I understand well, this relaxes the conditions under which a new set of inline ranges would be created to avoid last week's issue. So here a potential new edge case would now be that we would be _too_ relaxed and would consider two unrelated text nodes as part of the same text node? Are we confident this cannot happen?

The previous hot fix is exactly what did you mention is relaxes conditions by [removing .trim()](https://github.com/oat-sa/tao-core-ui-fe/commit/dac47c2a077e0f02f3334e45cd2e887a7c305319#diff-f5b8b9a29a8cbafec11f1a7c94b2ee014ca54fade83742049614bae8cd5950eaL685)
So I revert that changes here since they introduced new issues

> The `isText()` check is actually already performed inside of the `isHotNode() > isWrappable()` helper, can you explain why it needs to be performed alone apart from the other checks? (especially without the `isBlacklisted()` check)? Could we create the relevant test cases maybe?

The difference between `isText()` and `isWrappable()` is the last one skips nodes with whitespaces.
Since the whitespace were skipped by `isWrappable()` instead of setting **inline range** we get into that block for setting non-inline ranges - [R508-R514](https://github.com/oat-sa/tao-core-ui-fe/pull/262/files#diff-f5b8b9a29a8cbafec11f1a7c94b2ee014ca54fade83742049614bae8cd5950eaR508-R514)
As a result for the case where we have 2 highlights with whitespace between 
Screenshot 2021-03-25 at 15 12 52

The ranges were set incorrectly: 
Screenshot 2021-03-25 at 15 34 06
Instead of: 
Screenshot 2021-03-25 at 15 28 01
For now, I fixed by adding to the isolated node condition the rule that next node should NOT to be only the empty string as well.

👆It hepled and now we are come under the `isHotNode()` where we set `inlineRanges` as expected.
But here we have another issue - when we run recursively to build inline ranges we're again skipping the whitespace and it builds the ranges in the same way we had in the previous stage.
So I [fixed it as well by not skipping the whitespace](https://github.com/oat-sa/tao-core-ui-fe/pull/262/files#diff-f5b8b9a29a8cbafec11f1a7c94b2ee014ca54fade83742049614bae8cd5950eaR551-R554)

anshul119:
Works as expected, tested it with Andrii, live.

PR #264, opened 2021-03-26
fix/TR-187 add option for label text

meefox:
Related to https://oat-sa.atlassian.net/browse/TR-187
added options for changing the ok label for alert dialog.
similar to confirm dialog

to test
is a bit tricky you can test it together with https://github.com/oat-sa/extension-tao-test-runner-plugins/pull/183 in ie11 you can see cancel instead of ok button

### DIFF
--- a/src/dialog/alert.js
+++ b/src/dialog/alert.js
@@ -46,9 +46,16 @@ export default function dialogAlert(message, action, onCreateDialog, options) {
     options = _.defaults(options || {}, _options);
     dialogOptions = {
         message: message,
-        buttons: options.buttons.labels.ok || __('Ok'),
+        autoRender: true,
         autoDestroy: true,
-        autoRender: true
+        buttons: {
+            ok: {
+                id: 'ok',
+                type: 'info',
+                label: options.buttons.labels.ok || __('Ok'),
+                close: true
+            }
+        }
     };
     dlg = dialog(dialogOptions).on('create.dialog', function() {
         if (onCreateDialog) {
--- a/src/dialog/alert.js
+++ b/src/dialog/alert.js
@@ -13,28 +13,44 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2015-2019 (original work) Open Assessment Technologies SA ;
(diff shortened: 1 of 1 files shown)


src/dialog/alert.js
```diff
@@ -13,28 +13,44 @@
...
+            }
+        }
+    };
+    let dialogOptions;
+    let dlg;
+    options = _.defaults(options || {}, _options);
+    dialogOptions = {
         message: message,
-        buttons: 'ok',
+        buttons: options.buttons.labels.ok || __('Ok'),
```
jsconan, line 49: **nitpick:** This won't work as expected. When the button is supplied as a string, it is considered as an identifier, that will be use to link the built-in buttons. If the supplied id does not match a built-in button, it will generate a custom one, the label being the taken from the id. This might be ok if: the `close` flag was not missing (without it the dialog won't autoclose), and if the id was not taken from a translated label that will introduce regression when the id will be expected.

See: https://github.com/oat-sa/tao-core-ui-fe/blob/fbb967c48b8c68b1172a4b59c5b1d37e3933788e/src/dialog.js#L209-L222

**suggestion:** look at the implementation of the confirm dialog for a better approach: https://github.com/oat-sa/tao-core-ui-fe/blob/c11cec547adb9924adebd3d0a965c6c648e91e0e/src/dialog/confirm.js#L39-L74

jsconan (changes requested):
Unfortunately, while the main idea is good, this is not the correct approach. See my comment below.

Please also check the unit test, which is hanging because it is not closing anymore while clicking the ok button.

Run npm run test:keepAlive`, then open your browser on the displayed URL, and go to `/test/dialogAlert/test.html`.

image

src/dialog/alert.js
meefox: Thank you @jsconan.
I just test the label and click only in modern browsers.
But we should be careful here, I did a kind of port of confirm dialog, and then I had a security issue with fullscreen,
because of onOkBtn: fn()... is executed through `call`, it was blocked:
err

jsconan (approved):
Works as expected. A unit test would have been appreciated.

**Review Checklist**
- [ ] Changelog is updated according to changes (if applicable)

lecosson (approved):
**Review Checklist**
- [ ] Changelog is updated according to changes (if applicable)
image

PR #266, opened 2021-03-31
fix/MS-1071/Add tests for multi-color highlighter

andrey-ponamarev:
**Related to task:**  
https://oat-sa.atlassian.net/browse/MS-1071
https://oat-sa.atlassian.net/browse/MS-1076
https://oat-sa.atlassian.net/browse/MS-1077 
**Description:** Fix edge highlighter cases
**Unit tests cli:** `npm run test highlighter`

### DIFF
--- a/src/highlighter.js
+++ b/src/highlighter.js
@@ -13,7 +13,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2016 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2016-2021 (original work) Open Assessment Technologies SA;
  */
 /**
  * Highlighter helper: wraps every text node found within a Range object.
@@ -95,6 +95,14 @@ export default function (options) {
     }
 
     /**
+     * Returns all highlighted nodes
+     * @returns {JQuery<HTMLElement>}
+     */
+    function getHighlightedNodes() {
+        return $(containerSelector).find(`.${highlightingClasses.join(',.')}`);
+    }
+
+    /**
      * Highlight all text nodes within each given range
      * @param {Range[]} ranges - array of ranges to highlight, may be given by the helper selector.getAllRanges()
      */
@@ -106,6 +114,7 @@ export default function (options) {
                 currentGroupId = getAvailableGroupId();
 
                 // easy peasy: highlighting a plain text without any DOM nodes
+                // NOTE: The condition checks the whole node content and not a selected content in a given range, that allows to wrap whitespace
                 if (
                     isWrappable(range.commonAncestorContainer) &&
                     !isWrappingNode(range.commonAncestorContainer.parentNode)
@@ -148,6 +157,7 @@ export default function (options) {
             isWrapping = false;
             reindexGroups(getContainer());
             mergeAdjacentWrappingNodes(getContainer());
+            unWrapEmptyHighlights();
         });
 
         if (options.clearOnClick) {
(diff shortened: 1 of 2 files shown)


no-chris (approved):

test/highlighter/test.js
```diff
@@ -1050,4 +1069,505 @@ define(['jquery', 'lodash', 'ui/highlighter'], function($, _, highlighterFactory
 
         assert.equal($(fixtureContainer).find('.hl').length, 0, 'Third higlight has been discarded');
     });
+
+    QUnit.module('highlighter with multi colors');
+
+    QUnit.cases
+        .init([
```
adriloma, line 1076: To improve readability I would move this huge array up into as a const, like is done with `highlightRangeData`. Or even better move both to a different file and import them from there.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.22.9",
+    "version": "1.22.10",
```
adriloma, line 3: This is no longer needed, correct?

adriloma (approved):
Code review only. Just a minor comment regarding test.

test/highlighter/test.js
andrey-ponamarev: I don’t think to decouple test cases from the test it's a good idea and will improve readability

package.json
andrey-ponamarev: The bump version is required since it's `npm` package.

test/highlighter/test.js
jsconan: **thought:** Indeed, the data fixtures could have been moved to an external JSON resource, and the dynamic code  could have been generated based on properties. Anyway, this is not a big issue.

jsconan (approved):
It looks good. But I would encourage you to revert the version bump, unless it relates to a backport.

The unit tests are failing globally, but this is not due to your PR. The feature itself works well (`npm run test highlighter` or` npm run test:keepAlive` then open the browser)

**Review Checklist**
- [ ] Changelog is updated according to changes (if applicable)

src/highlighter.js
```diff
@@ -148,6 +157,7 @@ export default function (options) {
             isWrapping = false;
             reindexGroups(getContainer());
             mergeAdjacentWrappingNodes(getContainer());
+            unWrapEmptyHighlights();
```
no-chris, line 160: 1. would that make sense to do this before all the other post-processing steps? (`reindexGroups` and `mergeAdjacentWrappingNodes`)?
2. for consistency shouldn't we pass the container in the same fashion (`getContainer()`) rather than referencing directly `containerSelector` in `getHighlightedNodes`?

no-chris (commented):
Did not tested it, just CR. I only have a small comment

src/highlighter.js
andrey-ponamarev: 1. This process should be run right after `mergeAdjacentWrappingNodes`. 
It is possible to get a case when group IDs won't be in numerical sequence(is it a reason why did you ask?), but to be honest I still don’t understand the role of group IDs and I will be really appreciated if you can help me to figure it out.
2. At the beginning, I've started implementation from passing container in the way as we did for `reindexGroups`  and `mergeAdjacentWrappingNodes `, but then I changed it for multiple reasons.
The most important of that, the function will not run recursive(as other processes) that's why there is no sense to pass any parameters and event if the element is going to be pass as parameter, it's required to be wrap in jQuery once again, so that the call will be similar to:
```javascript
$($(containerSelector).get(0)).find(
```
But maybe it makes sense for caching the `getContainer()` and pass it to the post-proccesors functions 🤔

no-chris: I think the purpose of the group Id was to identify highlights nodes that are visually part of the same highlight. 
I don't have a specific error case in mind, I was just wondering if this could have an impact.

PR #367, opened 2021-09-30
Fix/TRN 1140/Highlighter: take blacklisting into account when building highlight index

oatymart:
https://oat-sa.atlassian.net/browse/TRN-1140

Small change needed to make NUI highlighter work with passages embedded in items.
Blacklisted nodes were not taken into account when building the highlights model.

TODO: unit test

### DIFF
--- a/src/highlighter.js
+++ b/src/highlighter.js
@@ -796,14 +796,17 @@ export default function (options) {
      */
     /**
      * For `keepEmptyNodes` option, creates data model of highlights.
-     * Additionally returns array of hilghlight nodes. Traverses DOM tree.
+     * Additionally returns array of highlight nodes. Traverses DOM tree.
      * @param {Node} rootNode
      * @returns {BuildModelResultKeepEmpty|null} result
      */
     function buildHighlightModelKeepEmpty(rootNode) {
         const classNames = options.colors ? Object.values(options.colors) : [className];
         const wrapperNodesSelector = classNames.map(cls => containerSelector + ' .' + cls).join(', ');
-        const wrapperNodes = document.querySelectorAll(wrapperNodesSelector);
+        const wrapperNodes = Array.from(document.querySelectorAll(wrapperNodesSelector))
+            .filter(node => {
+                return node.childNodes.length && !isBlacklisted(node.childNodes[0])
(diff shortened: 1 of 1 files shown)


src/highlighter.js
```diff
@@ -796,14 +796,17 @@ export default function (options) {
...
      * @param {Node} rootNode
      * @returns {BuildModelResultKeepEmpty|null} result
      */
     function buildHighlightModelKeepEmpty(rootNode) {
         const classNames = options.colors ? Object.values(options.colors) : [className];
         const wrapperNodesSelector = classNames.map(cls => containerSelector + ' .' + cls).join(', ');
-        const wrapperNodes = document.querySelectorAll(wrapperNodesSelector);
+        const wrapperNodes = Array.from(document.querySelectorAll(wrapperNodesSelector))
+            .filter(node => {
+                return node.childNodes.length && !isBlacklisted(node.childNodes[0])
```
olga-kulish, line 808: Question: isn't `isBlacklisted(node)` enough - why do you check inner text node instead of highlight span?

olga-kulish (commented):
Change looks correct

src/highlighter.js
oatymart: Yes - seems I had a mistake elsewhere before.

svoronuk (commented):
All changes looks correct. One small question.

src/highlighter.js
```diff
@@ -242,6 +244,17 @@ export default function (options) {
             }
             currentNode = childNodes[i];
 
+            if (isBlacklisted(currentNode)) {
+                if (isElement(currentNode)) {
```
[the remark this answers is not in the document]
oatymart: I made a bug here (my `continue` statement caused the final `break` block to be skipped), so I had to make another change. The resulting diff looks more complicated than it is, here's the high level:
```
if (isBlacklisted(currentNode)) {
    if (isElement(currentNode)) {
        //go deeper in case a descendant of the current blacklisted is whitelisted
        wrapTextNodesInRange(currentNode, rangeInfos);
    }
} else {
    // unchanged code, 75 lines
}
// end wrapping?
```

PR #368, opened 2021-10-01
CUNO-1853 / ADF-695: User without access to assets can save images trough the item

clagosarias:
**Jira issue**
https://oat-sa.atlassian.net/browse/ADF-695

**Related PR**
https://github.com/oat-sa/tao-core/pull/3151

**AC**
- Preselect item gallery when user doesn't have access to Assets.
- Preselect Assets when user does have access.
- Make "Add file" button appear without having to click a folder. (If permissions allow it).

**How to test**
- Make access to Assets forbidden for user. (By restricting access only to Global Manager role)(make it recursive)
- Login with user and author an item
- Try to insert media/image

PR #372, opened 2021-10-06
Backport/adf 695/item gallery default

clagosarias:
**Related task**
https://oat-sa.atlassian.net/browse/ADF-695

**Backport of the fix**
https://github.com/oat-sa/tao-core-ui-fe/pull/368

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.34.3",
+    "version": "1.34.4",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-core-ui",
-    "version": "1.34.3",
+    "version": "1.34.4",
     "displayName": "TAO Core UI",
     "description": "UI libraries of TAO",
     "scripts": {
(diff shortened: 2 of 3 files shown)


404passionFruit:
Shouldn't the target branch be different? 

> For npm libraries (ex. @oat-sa/tao-core-sdk, @oat-sa/tao-test-runner-qti) you can't use 4th number in the version, see https://semver.org/
> For the version 0.4.1 you may create 0.4.1-1

https://oat-sa.atlassian.net/wiki/spaces/OAT/pages/114885777/How+to+backport

Is this practice up to date?

clagosarias:
> Shouldn't the target branch be different?
> 
> > For npm libraries (ex. @oat-sa/tao-core-sdk, @oat-sa/tao-test-runner-qti) you can't use 4th number in the version, see https://semver.org/
> > For the version 0.4.1 you may create 0.4.1-1
> 
> https://oat-sa.atlassian.net/wiki/spaces/OAT/pages/114885777/How+to+backport
> 
> Is this practice up to date?

@404passionFruit I think that is in the context of the target version already existing. In which case you must use the 4th number, but it is not the case.

> Basically, you'll have to create a new version from the current one. Let's assume the current version to patch is 20.4.4. You'll have to create a version above, and you can face one of these two cases:
> 
> - the version 20.4.5 does not exist, so you may create it without any problem
> -  the version 20.4.5 already exists, so you have to create an intermediate one, like 20.4.4.1 or 20.4.4.2, etc.

404passionFruit (approved):
Good job!

PR #373, opened 2021-10-15
Feature / ADF-710 [GREP] - prevent changes of property list values and property type

clagosarias:
**Ticket**
https://oat-sa.atlassian.net/browse/ADF-710

**Related PR**
https://github.com/oat-sa/tao-core/pull/3172

**Description**
Had to implement styling for new `form_disabled_message` in manage schema form, so I took the opportunity to move styling from tao-core extension's `layout.scss` into tao-core-ui. 
Created a new scss file for it and applied already existing styling for secondary-properties


scss/inc/_secondary-properties.scss
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        > div {
+            display: flex;
+            align-items: center;
+            justify-content: space-between;
+            flex-wrap: wrap;
+
+            &.disabled {
+                background: transparent!important;
```
404passionFruit, line 52: I suppose there is no way to avoid `!important`, right?

404passionFruit (approved):
Good job!

PR #392, opened 2021-11-26
Fix/TR-2846/Improve stalled detection

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-2846

In the context of the pre-caching features, the media player is not always able to properly detect a stalled media. It may happen that the state is set even if the media is not actually stalled.

This is especially true with Safari since the MIME type detection was inaccurate with cached audio (the pre-caching loader is caching the data in a blob). The MIME type detection was reworked to first rely on the supplied MIME type if any, and then only fallback to server side detection if missing.

Some additional fixes have been made as well:
- the look and feel for the "reload" state for audio player
- the support detection for WebM media in Safari

### DIFF
--- a/src/mediaplayer.js
+++ b/src/mediaplayer.js
@@ -189,9 +189,14 @@ const isResponsiveSize = sizeProps => {
 /**
  * Builds a media player instance
  * @param {Object} config
- * @param {String} config.type - The type of media to play
- * @param {String|Array} config.url - The URL to the media
- * @param {String} [config.mimeType] - The MIME type of the media
+ * @param {String} config.type - The type of media to play, say `audio`, `video`, or `youtube`. The default is `video`.
+ * It might also contain the MIME type of the media as a shorthand.
+ * @param {String|Array} [config.url] - The URL to the media. If several media are proposed as alternatives,
+ * please look at the `sources` option instead.
+ * @param {String} [config.mimeType] - The MIME type of the media. If omitted, the player will try to extract it
+ * from the `type` property, otherwise it will request the server to get the content-type.
+ * @param {Array} [config.sources] - A list of URL if several media can be proposed. Each entry may be either a
+ * string (single URL), or an object containing both the URL and the MIME type ({src: string, type: string}).
  * @param {String|jQuery|HTMLElement} [config.renderTo] - An optional container in which renders the player
  * @param {Boolean} [config.canSeek] - The player allows to reach an arbitrary position within the media using the duration bar
  * @param {Boolean} [config.loop] - The media will be played continuously
--- a/src/mediaplayer.js
+++ b/src/mediaplayer.js
@@ -191,6 +191,7 @@ const isResponsiveSize = sizeProps => {
  * @param {Object} config
  * @param {String} config.type - The type of media to play
  * @param {String|Array} config.url - The URL to the media
+ * @param {String} [config.mimeType] - The MIME type of the media
  * @param {String|jQuery|HTMLElement} [config.renderTo] - An optional container in which renders the player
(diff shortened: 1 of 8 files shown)


oatymart (commented):
First, I can confirm that the mime type setting in Safari is fixed, and the visual improvements are good.

Second, regarding the webM video - it loads correctly in Safari 14, but not in 13.1:
Screenshot 2021-11-29 at 14 34 41
However, maybe it was like this before?

Finally, about the stalled detection false positive in Windows Chrome - I also haven't been able to reproduce it in Browserstack, or any other browser locally. I've experimented with passing different timeout values for the initial `detectStalledNetwork` call and the later one, but with no useful outcome so far.

jsconan:
Indeed, Safari 13 does not support WebM, at least not all codecs, and trying to open the video in a new tab does not trigger the player in Safari 13. In Safari 14 it does.

Actually, you can see that Safari 13 has a partial support of this container: https://caniuse.com/?search=webm

olga-kulish (commented):
Changes look good.

For '_Disconnection detected right from the start_', I still couldn't reproduce it. It happened only once when request was **really** hanging for a couple of minutes. Actually, I also have several times run into situation where page was hanging while moving between items (on https://beosa01dee.selor.taocloud.org/). So I had to refresh the page to continue. Unfortunately, I didn't check which request exactly was hanging then.


For '_Slow 3G case_', I added a comment in TR-2896 https://oat-sa.atlassian.net/browse/TR-2896?focusedCommentId=163029 . As I don't have a proof that increasing timeout is really helpful, I'm fine to both leave it as now or increase it.

src/mediaplayer/scss/player.scss
```diff
@@ -343,6 +343,32 @@ $controlsHeight: 36px;
...
+                        flex-direction: row;
+                        justify-content: center;
+
+                        &.reload {
+                            width: 100%;
+                            font-size: 20px;
+                            line-height: 20px;
+
+                            .message {
+                                font-size: 10px;
```
olga-kulish, line 363: Just a note: I assume this font size is fine for ux. It's very small, but I don't have suggestion for improvement.

src/mediaplayer.js
```diff
@@ -191,6 +191,7 @@ const isResponsiveSize = sizeProps => {
  * @param {Object} config
  * @param {String} config.type - The type of media to play
  * @param {String|Array} config.url - The URL to the media
+ * @param {String} [config.mimeType] - The MIME type of the media
```
olga-kulish, line 194: Do we really need a separate property? If you have a case in mind where `type` and `mimeType` are different, maybe add an example of their values in jsdoc here? 

[MediaInteraction uses `type` and doesn't set `mimeType` anyway.]

olga-kulish (commented):
Works well - type from itemData is applied to media html element, and stalled audio overlay doesn't go outside borders

src/mediaplayer.js
jsconan: The `type` is expected to be either `audio` or `video`. Nothing oblige to make it a MIME type, and this is a shorthand to also have the MIME type inside it. Usually, the MIME type is supplied together with the media source, in the `source` parameter. But the source can also be supplied through the `url` parameter. Since nothing ensure the type is having the MIME as well, we need to have a way to supplied it properly.

But yes, I agree, the documentation is minimalistic. I updated it.

src/mediaplayer/scss/player.scss
jsconan: Actually, it needs to fit in the audio player space. So not so many choices unfortunately.

olga-kulish (approved):
Well done!

oatymart (approved):
Changes have been tested, all good. Nice job on a tricky task!

PR #406, opened 2021-12-22
fix: TR-3163. Restore focusing functionality

svoronuk:
Related to: https://oat-sa.atlassian.net/browse/TR-3163

This style change fix problem with breaking focusing flow for pages with checkboxes.
Root cause of this - `visibility:hidden` style do not allow focus element. This style was added in terms of fixing this: https://oat-sa.atlassian.net/browse/TR-916. After testing with author we decided the changing `opacity` instead of `visibility` is a good compromise for both tickets.

How to test:
1. switch repo `tao-core` to this PR: https://github.com/oat-sa/tao-core/pull/3295
2. use this repo
3. run test with choice interaction
4. move focus using keyboard
5. be sure focus jumps and console without errors


scss/inc/_forms.scss
```diff
@@ -412,8 +412,6 @@ label {
     }
     input[type="radio"],
     input[type="checkbox"] {
-        visibility: hidden;
-        pointer-events: none;
```
olga-kulish, line 416: It's not so harmless:

Screenshot_5

vs

Screenshot_6

Note that before `visibility:hidden`, it used offscreen `left` position: https://github.com/oat-sa/tao-core-ui-fe/commit/17b92c7f5098b9087c41d45267d900d9e6163013#diff-3bc1145cd4a253910fe0ef19c9411907c213932fc888152ff7dafefa924c126d
Maybe you should keep that solution (why did commit author remove it?).

olga-kulish (commented):
Thank you for investigating this!

Also, TR-3163 is a technical ticket, but this change affects users and should be QAed. Do you intend to write a comment in the current ticket or create a linked bug? Maybe we could also leave unit tests failing...

scss/inc/_forms.scss
svoronuk: Actually solution which we discussed is a little bit wider. I just cut one property, because i found that it duplicates in other css rule. I tested it on some pages in my test and it works. I will revert this cut property and test it once more.

About the testing you are right. Typically we do not test technical tickets, but it is not a rule, it is because no way to test in QA. Here other situation. I ask our QA, which way they prefer: separate ticket or do it in terms of this.

svoronuk: I found one more issue in authoring and fix this, but in any case i will create separate ticket and will continue on it.

PR #407, opened 2021-12-23
fix: TR-3292. Restore focusing functionality

svoronuk:
Related to: https://oat-sa.atlassian.net/browse/TR-3292

This style change fix problem with breaking focusing flow for pages with checkboxes.
Root cause of this - `visibility:hidden` style do not allow focus element. This style was added in terms of fixing this: https://oat-sa.atlassian.net/browse/TR-916. After testing with author we decided the changing `opacity` instead of `visibility` is a good compromise for both tickets.

How to test: test it in terms of this PR https://github.com/oat-sa/tao-core/pull/3296


scss/inc/_forms.scss
```diff
@@ -410,11 +410,14 @@ label {
...
+    input[type="radio"]:disabled,
+    input[type="checkbox"]:disabled {
+        opacity: 0;
+    }
     input[type="radio"],
     input[type="checkbox"] {
-        visibility: hidden;
-        pointer-events: none;
         position: absolute;
+        opacity: 0;
```
olga-kulish, line 420: I'm still not 100% happy with this since it still takes space and is clickable (yes, an edge case)
Maybe see `visually-hidden` from LDS https://github.com/oat-sa/live-design-system/blob/master/packages/identity/css/mixins/_statuses.css#L43 ?
What do you think?

https://user-images.githubusercontent.com/38751932/147256403-2b4f1884-1598-408f-a118-a8036f0f5da1.mp4

svoronuk: Yeah. Looks weird. I'll change this.

svoronuk: Convert approach to LDS-similar, as you suggested.

scss/inc/_forms.scss
```diff
@@ -410,11 +410,17 @@ label {
...
-        visibility: hidden;
-        pointer-events: none;
+        /* based on "visually-hidden" mixin in LDS for accessibility goals */
         position: absolute;
+        width: 1px;
+        height: 1px;
+        overflow: hidden;
+        clip: rect(1px, 1px, 1px, 1px);
+        margin: 0;
+        padding: 0;
```
oatymart, line 423: It's a good idea. I would even propose defining it as a `visuallyHidden` mixin in `scss/inc/_functions.scss`, because we've seen how useful the pattern was in Solar.

scss/inc/_functions.scss
```diff
@@ -270,3 +270,13 @@ Usage:
 @mixin disableSelect() {
     @include vendor-prefix(user-select, none, property);
 }
+
+/* based on "visually-hidden" mixin in LDS for accessibility goals */
+@mixin visuallyHidden() {
```
oatymart, line 275: I think `position: absolute;` is still required, to remove it from the flow (so that neighbouring elements would not be moved by 1px). We have it in the LDS, and the [original article](https://snook.ca/archives/html_and_css/hiding-content-for-accessibility) has it too.

svoronuk: Agree. Very strange, but i copied this with position property. In any case i'll fix this.

olga-kulish (approved):
Works well.
Of course, I didn't check all possible use-cases of this common component)

- [ ] New code is respecting code style rules
- [ ] New code is respecting best practices

shaveko (approved):
Looks good to me

oatymart (approved):
The change is correct in terms of removing the (bad for accessibility) `visibility: hidden` and replacing it with the most appropriate modern CSS.

Fixes the issue and shouldn't have any negative side effects, although I did not check every possible UI situation involving checkbox/radio.

jsconan (approved):
Works well!

**Review Checklist**
- [x] Commits are atomic

PR #408, opened 2022-01-03
fix/TR-3280/header-logo-support-rtl

marpesia:
**Related to:** [TR-3280](https://oat-sa.atlassian.net/browse/TR-3280)

**Description:**
On RTL mode the logo on the Header keeps on the left. There is a CSS class that forces a `float: left`.
I've tracked the CSS class but I don't know the reason to have it. In order to avoid regression, I just disable the floating effect on RTL mode.

**Changes:**

- Add CSS rule to disable `float: left` effect on RTL mode.

### DIFF
--- a/scss/inc/_base.scss
+++ b/scss/inc/_base.scss
@@ -382,6 +382,12 @@ body {
         float: left;
     }
 
+    body[dir="rtl"] {
+        & .content-wrap>header>.lft {
+            float: none;
+        }
+    }
+
     .rgt,
     .wrap-right {
         display: inline;


bziondik (approved):
- [ ]  New code is covered by tests (if applicable)
- [ ]  Tests are running successfully (old and new ones) on my local machine (if applicable)
- [x]  Feature is working correctly on playground (if applicable)

krampstudio (changes requested):
IMO, this is not the right place since `.content-wrap` is coming from the layout. This `_base.scss` seems to provide some base classes, we shouldn't have layout rule inside it.

marpesia:
> IMO, this is not the right place since `.content-wrap` is coming from the layout. This `_base.scss` seems to provide some base classes, we shouldn't have layout rule inside it.

Yes, that was my first idea, but it is working once the user is logged and this RTL CSS rule is here and not in the `_base.scss` so I thought it is better to keep RTL rules together. WDYT?

PR #471, opened 2022-07-08
feat/AUT-2332 Add language to editor

meefox:
Related to https://oat-sa.atlassian.net/browse/AUT-2332

Main description https://github.com/oat-sa/tao-core/pull/3575

**Description:**
Source changes for CKE skin modifications, please refer to this PR related to CKE editor changes.

**Changes:**
* Added language plugin to the toolbar, adapted tao skin for the plugin.

Part 4/6

### DIFF
--- a/scss/ckeditor/skins/tao/scss/editor.scss
+++ b/scss/ckeditor/skins/tao/scss/editor.scss
@@ -109,7 +109,3 @@ legend.cke_voice_label {
   text-shadow: none;
   color: $textColor;
 }
-
-.cke_focus span[dir] {
-  background-color: $uiReviewPanelBg;
-}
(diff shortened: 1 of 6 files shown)


jsconan (commented):
Quick code review only.
It looks good so far.

I only have a few remarks, mainly about the hardcoded colors. I saw there are several places where they are inlined in the existing code, so maybe was is not made using the core variables, but it looks a bit nasty to me. It would be better to use variables, but if this stylesheet is not ready to use them it is not a blocker.

Then another one regarding the way to activate the plugin. I wonder if we should not activate from another place to give more control over it with respect to the show/hide features.

shaveko (approved):
It looks nice codewise. Checked functionally as whole on test env. Good job!

PR #494, opened 2022-09-20
Feature/adf 1197/refactor search modal requests

jsconan:
### Summary

A few changes:
- Preliminary changes to make the search query more agnostic.
- Support for sorting the columns with respect to the new columns configuration.
- Preliminary support for the columns selection (in the state, all available columns are displayed).

Note: as a small improvement, this PR also fixes the port number for the unit tests runner. It is now 5400 instead a random port.

### Details

Move the hardcoded routes to the config. There are still changes needed on the consumer side, the config now relies on default values.

Modularise the search preparation, using promise chains.
Wrong concurrency management was also fixed (useless throttle of nested function).

The [data model ](https://github.com/oat-sa/tao-core-ui-fe/pull/494/files#diff-931d32bcbcc0445525c09cb03af3e8d008e666548a6a8b32d94e9938d07a8730R420)for the datatable is now coming from the new API (the `settings` section of the `searchParams` response). Support is still present for the old `model` property, but it could be removed in the end.

Support for a sort identifier different from the column identifier is also added.

What is remaining: wire the column selection to the data model. For the time being, all available columns are used.

### How to test
-  run the unit tests
    ```sh
    npm run test:keepAlive
    ```
    Open the browser at (replace PORT_NUMBER):
    - `http://127.0.0.1:PORT_NUMBER/test/searchModal/advancedSearch/test.html`
    - `http://127.0.0.1:PORT_NUMBER/test/searchModal/test.html`
- checkout the branch in `tao/views/node_module/@oat-sa/tao-core-ui`, and check the advanced search in TAO (you will need to also checkout the integration branches in the other repositories)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "description": "UI libraries of TAO",
     "scripts": {
         "test": "npx qunit-testrunner",
-        "test:keepAlive": "npx qunit-testrunner --keepalive",
+        "test:keepAlive": "npx qunit-testrunner --keepalive --port 5400",
         "test:cov": "npm run build:cov && npx qunit-testrunner --cov",
         "test:dev": "run-p test:keepAlive build:watch",
         "coverage": "nyc report",
--- a/src/datatable/tpl/layout.tpl
+++ b/src/datatable/tpl/layout.tpl
@@ -61,7 +61,7 @@
                         <div
                             {{#if sortable}}
                                 class="sortable"
-                                data-sort-by="{{id}}"
+                                data-sort-by="{{#if sortId}}{{sortId}}{{else}}{{id}}{{/if}}"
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
     "description": "UI libraries of TAO",
     "scripts": {
         "test": "npx qunit-testrunner",
-        "test:keepAlive": "npx qunit-testrunner --keepalive",
+        "test:keepAlive": "npx qunit-testrunner --keepalive --port 5400",
```
shaveko, line 8: Nice, that port change bothered a lot

shaveko (approved):
Looks nice and works well. Did not find any issue neither in code nor in functionality

PR #495, opened 2022-09-22
Feat/adf 1193/add property selection to search modal

shaveko:
related to https://oat-sa.atlassian.net/browse/ADF-1193

### Description 

Use `propertySelector` component in `searchModal`

### How to test

`npm run test:keepAlive` and play with `searchModal` component


src/searchModal.js
```diff
@@ -419,13 +441,51 @@ export default function searchModalFactory(config) {
      */
     function buildDataModel(data) {
         if (data.settings) {
+            //save availableColumns to memory
+            availableColumns = data.settings.availableColumns;
             // @todo: use the selected columns instead. It can use a promise as it takes place insise a promise chain
-            data.model = columnsToModel(data.settings.availableColumns);
+            data.model = columnsToModel(availableColumns);
```
jsconan, line 447: **chore:** As it seems the columns are refined after, and not before as it was foreseen, you could remove the coment above.

**thought:** Looking at the code, I think the function `buildDataModel` should be removed for the benefit of `filterSelectedColumns` which should do the job. Having the model refined, then filtered looks a waste of resources. I'd rather do the filtering then the refining, or even both at once.

src/searchModal.js
```diff
@@ -419,13 +441,51 @@ export default function searchModalFactory(config) {
...
+    }
 
-        return data;
+    /**
+     * Filters datatble model based on stored selected columns
+     * @param {Object} data data containing available columns and model for datatable
+     * @returns {Promise} promise which resolves with filtered data.model
+     */
+    function filterSelectedColumns(data) {
+        console.log(data);
```
jsconan, line 463: **chore:** Forgotten debug line

src/searchModal.js
```diff
@@ -419,13 +441,51 @@ export default function searchModalFactory(config) {
...
+        return selectedColumnsStore
+            .getItem(getClassFilterUri())
+            .then(storedSelectedColumnIds => {
+                //let columnsToDisplay;
+                if (storedSelectedColumnIds && storedSelectedColumnIds.length) {
+                    selectedColumns = [...storedSelectedColumnIds];
+                    //columnsToDisplay = data.settings.availableColumns.filter(column =>
+                    // selectedColumns.includes(column.id)
+                    //);
```
jsconan, line 472: **chore:** Dead code. Was it expected to remove it, or is it a temporary disabling that was not reverted?

src/searchModal.js
```diff
@@ -419,13 +441,51 @@ export default function searchModalFactory(config) {
...
+                    }, []);
+                }
+
+                data.model = data.model.filter(column => selectedColumns.includes(column.id));
+
+                return data;
+            })
+            .catch(e => {
+                instance.trigger('error', e);
+                reject(new Error('Error getting selected columns'));
```
jsconan, line 487: **thought:** Is there a need to trigger this error? It does not seem to be caught. It would produce unresolved promise errors, but actually, there is no wrapping promise. Is it a dead code?

src/searchModal.js
```diff
@@ -490,6 +563,52 @@ export default function searchModalFactory(config) {
...
+                }
+                propertySelectorInstance.hide();
+            });
+        } else {
+            if (propertySelectorInstanceHidden) {
+                propertySelectorInstance.setData({ selected: selectedColumns });
+                propertySelectorInstance.show();
+            } else {
+                propertySelectorInstance.hide();
+            }
```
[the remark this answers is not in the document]
shaveko: Thanks, done

src/searchModal.js
```diff
@@ -490,6 +563,52 @@ export default function searchModalFactory(config) {
...
+            });
+            propertySelectorInstance.on('cancel', propertySelectorInstance.hide);
+            propertySelectorInstance.on('hide', () => {
+                propertySelectorInstanceHidden = true;
+            });
+            propertySelectorInstance.on('show', () => {
+                propertySelectorInstanceHidden = false;
+            });
+
+            propertySelectorInstance.on('select', e => {
```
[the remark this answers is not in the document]
shaveko: Got it, refactored, thanks

src/searchModal.js
```diff
@@ -419,13 +433,50 @@ export default function searchModalFactory(config) {
...
+                }
 
-        return data;
+                data.model = data.model.filter(column => selectedColumns.includes(column.id));
+
+                return data;
+            })
+            .catch(e => {
+                instance.trigger('error', e);
+                reject(new Error('Error getting selected columns'));
```
jsconan, line 474: **suggestion:** Actually, the idea of making the error more accurate is good. But then, it should go with error wrapping. The returned promise is caught later, with another trigger, so no need to double it.
```suggestion
                return Promise.reject(new Error('Error getting selected columns', { cause: e }));
```

src/searchModal.js
```diff
@@ -419,13 +433,50 @@ export default function searchModalFactory(config) {
...
+                    selectedColumns = [...storedSelectedColumnIds];
+                    //columnsToDisplay = data.settings.availableColumns.filter(column =>
+                    // selectedColumns.includes(column.id)
+                    //);
+                } else {
+                    selectedColumns = data.settings.availableColumns.reduce((acc, column) => {
+                        if (column.default) {
+                            acc.push(column.id);
+                        }
+                    }, []);
```
jsconan, line 465: **issue:** The accumulator is never returned, and the browser will issue a runtime error: `TypeError: can't access property "push", acc is undefined`

**suggestion:**
```suggestion
                    selectedColumns = data.settings.availableColumns.reduce((acc, column) => {
                        if (column.default) {
                            acc.push(column.id);
                        }
                        return ACC;
                    }, []);
```

src/searchModal.js
```diff
@@ -468,12 +528,25 @@ export default function searchModalFactory(config) {
...
     /**
      * Triggered on load.datatable event, it updates searchStore and manages possible exceptions
      * @param {object} e - load.datatable event
      * @param {object} dataset - datatable dataset
      */
     function searchResultsLoaded(e, dataset) {
+        const $actionsHeader = $('th.actions', $container);
+        const $manageColumnsBtn = $(propertySelectButtonTpl());
+        $actionsHeader.append($manageColumnsBtn);
+        $manageColumnsBtn.on('click', handleManageColumnsBtnClick);
```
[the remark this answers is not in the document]
jsconan: Well, it is not an issue after all. Actually, as the whole datatable is written again and again each time it is updated, this is the correct approach to add the button each time.
I'll resolve the comment.

bziondik (commented):
One suggestion: clean filter in component after close by button, to have it without filter after reopen or have cross icon on the right to clean search input 
image

src/searchModal.js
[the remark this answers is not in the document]
shaveko: Ok, thank you for this catch! Now I've bound to `create.datatable`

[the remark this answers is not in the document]
shaveko: Yes, right, removed

src/searchModal.js
shaveko: Yes, right, removed

src/searchModal.js
shaveko: Yes, I've forgot to remove while refactored

src/searchModal.js
shaveko: actually `reject` is not defined here

src/searchModal.js
shaveko: yes, exactly. But I just did remove that reject code

src/searchModal.js
shaveko: Yes, right. I wonder how I've missed that..

shaveko:
> One suggestion: clean filter in component after close by button, to have it without filter after reopen or have cross icon on the right to clean search input image

Nice Idea, thanks!

jsconan (commented):
It looks very nice now. Good job.

2 (small) last things:
- the `manage columns` link disappear when the table refreshes (sort or paginate). I apologize for the confusion. I is due to a comment I made earlier about the life cycle of the table but I was wrong and I added an erratum about it. Unfortunately, I think you missed it and you did the unnecessary fix.
- the style of the `manage columns` is not totally the same as the mockup was showing
    - the implemented button: image
    - the mockup: image

jsconan (approved):
It looks nice and it works great. Good job!

One last suggestion, but it can be refined later: disable the user selection on the `manage columns` link.
image

PR #509, opened 2022-10-06
Fix/ADF-1220/broken layout

jsconan:
### Summary

Fix the broken layout issue after a clear.

### Details

Once the search results were cleared, the search did not work anymore. This was due to one of the last changes, for moving the manage columns link outside of the table. The layout was reworked, but the code was still targeting the wrong selector for clearing the content, breaking the layout when a new search is done.

This is fixed thanks to a small refactor for removing the bad design with multiple inline selectors used here and there.

### How to test
-  compile the code
    ```sh
    npm run build:all
    ```
-  run the unit tests
    ```sh
    npm run test:keepAlive
    ```
    Open the browser at:
    - http://127.0.0.1:5400/test/searchModal/advancedSearch/test.html
    - http://127.0.0.1:5400/test/searchModal/test.html
- checkout the branch in `tao/views/node_module/@oat-sa/tao-core-ui`, and check the advanced search in TAO (you will need to also checkout the integration branches in the other repositories)

### DIFF
--- a/src/searchModal.js
+++ b/src/searchModal.js
@@ -71,16 +71,11 @@ export default function searchModalFactory(config) {
     };
     // Private properties to be easily accessible by instance methods
     let $container = null;
-    let $searchInput = null;
-    let $searchButton = null;
-    let $clearButton = null;
+    let controls = {};
     let running = false;
     let searchStore = null;
     let selectedColumnsStore = null;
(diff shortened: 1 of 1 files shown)


shaveko (approved):
Looks good and works well. Thumbs up for `controls` refactoring

PR #528, opened 2022-11-15
fix/AUT-2673/inline-figure-img

marpesia:
**Related to:** [AUT-2673](https://oat-sa.atlassian.net/browse/AUT-2673)

**Parent Extension**
- https://github.com/oat-sa/tao-core/pull/3676

**Changes:**

- Trigger re-build Figure widger on inline alignment.
- Add param to Widget panel to show a message on the inline case that prevents the use of FigCaption

### DIFF
--- a/src/figure/FigureStateActive.js
+++ b/src/figure/FigureStateActive.js
@@ -102,7 +102,8 @@ const initForm = ({ widget, formElement, formTpl, mediaEditor, togglePlaceholder
             baseUrl: widget.options.baseUrl || '',
             src: imageElem.attr('src'),
             alt: imageElem.attr('alt'),
-            figcaption: figcaptionElem ? figcaptionElem.body() : ''
+            figcaption: figcaptionElem ? figcaptionElem.body() : '',
+            showFigure: widget.element.attr('showFigure')
         })
     );
 
--- a/src/figure/FigureStateActive.js
+++ b/src/figure/FigureStateActive.js
@@ -97,13 +97,14 @@ const formCallbacks = ({ widget, formElement, mediaEditor, togglePlaceholder }) 
 const initForm = ({ widget, formElement, formTpl, mediaEditor, togglePlaceholder }) => {
     const imageElem = getImageElement(widget);
     const figcaptionElem = getCaptionElement(widget);
+    const showFigure = widget.element.attr('showFigure');
     widget.$form.html(
         formTpl({
             baseUrl: widget.options.baseUrl || '',
             src: imageElem.attr('src'),
             alt: imageElem.attr('alt'),
             figcaption: figcaptionElem ? figcaptionElem.body() : '',
-            showFigure: widget.element.attr('showFigure')
+            showFigure: showFigure
         })
     );
 
--- a/src/figure/FigureStateActive.js
+++ b/src/figure/FigureStateActive.js
@@ -133,10 +133,9 @@ const initForm = ({ widget, formElement, formTpl, mediaEditor, togglePlaceholder
  * @param {Object} formTpl
  * @param {Object} formElement 
  * @param {Object} inlineHelper
- * @param {Object} htmlEditor
  * @returns 
  */
-export default function ({ stateFactory, ActiveState, formTpl, formElement, inlineHelper, htmlEditor }) {
+export default function ({ stateFactory, ActiveState, formTpl, formElement, inlineHelper }) {
     /**
      * media Editor instance if has been initialized
      * @type {null}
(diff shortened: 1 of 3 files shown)


src/figure/FigureStateActive.js
```diff
@@ -170,5 +173,14 @@ export default function (stateFactory, ActiveState, formTpl, formElement, inline
         }
     };
 
+    ImgStateActive.prototype.destroyEditor = function () {
+        const widget = this.widget;
+        const $editableContainer = widget.$original.parents('[data-qti-class="_container"]');
+        htmlEditor.buildEditor($editableContainer, {});
+        _.defer(() => {
+            htmlEditor.destroyEditor($editableContainer);
+        });
```
jsconan, line 182: **question:** Why building then destroying the editor immediately?

src/figure/FigureStateActive.js
```diff
@@ -125,7 +127,7 @@ const initForm = ({ widget, formElement, formTpl, mediaEditor, togglePlaceholder
     );
 };
 
-export default function (stateFactory, ActiveState, formTpl, formElement, inlineHelper) {
+export default function (stateFactory, ActiveState, formTpl, formElement, inlineHelper, htmlEditor) {
```
jsconan, line 130: **quibble:** The signatures becomes heavier and heavier. Maybe it is the right time to think about simplifying it. If the API is not used in too many places, it could be simplified with a config object, allowing to addition of more entries without breaking the signature.

jsconan (commented):
Code check only.
It looks okayish. But I have a few concerns :)

src/figure/FigureStateActive.js
marpesia: The CKEditor reads the Figure HTML tag, and if it is inside a block, it moves out of it.
To achieve this effect, I must rebuild the editor, but I don't want to have it active, so I destroy it afterwards. Otherwise, I will have an error in the XML structure.

src/figure/FigureStateActive.js
marpesia: Good point!

src/mediaEditor/plugins/mediaAlignment/helper.js
```diff
@@ -43,12 +43,17 @@ export const positionFloat = function positionFloat(widget, position) {
...
     // Update model
+    const prevClassName = widget.element.attr('class');
     if (className) {
         widget.element.attr('class', className);
     } else {
         widget.element.removeAttr('class');
     }
-
+    
+    if ((prevClassName || className.length) && prevClassName !== className) {
```
jsconan, line 53: **question:** Why test the value only for one variable and the length only for the other variable? Both are string, so it could be tested by value for both...

**suggestion:**
```suggestion
    if ((prevClassName || className) && prevClassName !== className) {
```

src/figure/FigureStateActive.js
```diff
@@ -125,7 +127,18 @@ const initForm = ({ widget, formElement, formTpl, mediaEditor, togglePlaceholder
...
+ * @param {Object} config
+ * @param {Object} config.stateFactory
+ * @param {Object} config.ActiveState
+ * @param {Object} config.formTpl
+ * @param {Object} config.formElement 
+ * @param {Object} config.inlineHelper
+ * @param {Object} config.htmlEditor
+ * @returns 
+ */
+export default function (config) {
```
jsconan, line 141: **suggestion:** It could also have been turned into a restructured declaration as mentioned in [the coding guide](https://github.com/oat-sa/taohub-articles/blob/master/articles/Frontend/Architecture/coding-guide.md#es2015-style)
```suggestion
export default function ({ stateFactory, ActiveState, formTpl, formElement, inlineHelper, htmlEditor }) {
```

If some parameters are mandatory and need to be given anyway, it could also be split into groups:
```suggestion
export default function (stateFactory, ActiveState, { formTpl, formElement, inlineHelper, htmlEditor }) {
```

jsconan (approved):
Code check only.

It looks good.

bziondik (approved):
- [ ]  New code is covered by tests (if applicable)
- [x]  Tests are running successfully (old and new ones) on my local machine (if applicable)
- [x]  Feature is working correctly on playground (if applicable)

Silvia-loza (approved):
Great job! :tada::tada::tada:

- [X] New code is respecting code style rules
- [X] New code is respecting best practices
- [X]  Pull request's target is not` master`
- [X] Feature is working correctly on kitchen env (if applicable)
- [X] Acceptance criteria are respected
- [X] Pull request title and description are meaningful

PR #555, opened 2023-04-14
Feature/adf 1424/acl section on destination selector

rgomeztao:
coming from [https://oat-sa.atlassian.net/browse/ADF-1424](https://oat-sa.atlassian.net/browse/ADF-1424)

changelog:
* adds acl radio buttons to destination selector

### DIFF
--- a/src/destination/selector.js
+++ b/src/destination/selector.js
@@ -96,8 +96,8 @@ export default function destinationSelectorFactory($container, config) {
             var self = this;
             var $component = this.getElement();
 
-            $('.acl-mode-radio').change(function () {
-                $('.acl-mode-radio').not(this).prop('checked', false);
+            $('.acl-mode-radio', $component).change(function () {
+                $('.acl-mode-radio', $component).not(this).prop('checked', false);
             });
 
             /**
@@ -116,7 +116,7 @@ export default function destinationSelectorFactory($container, config) {
                 }
             };
 
-            const getSelectedACLTransferMode = () => $('input.acl-mode-radio:checked').val();
+            const getSelectedACLTransferMode = () => $('input.acl-mode-radio:checked', $component).val();
 
             if (this.config.taskQueue) {
                 this.taskCreationButton = taskCreationButtonFactory({
--- a/src/destination/selector.js
+++ b/src/destination/selector.js
@@ -36,7 +36,8 @@ var defaultConfig = {
     title: __('Copy to'),
     description: __('Select a destination'),
     actionName: __('Copy'),
-    icon: 'copy'
+    icon: 'copy',
+    showACL: false
 };
 
 /**
@@ -86,6 +87,7 @@ export default function destinationSelectorFactory($container, config) {
     )
         .setTemplate(selectorTpl)
         .on('init', function() {
+            this.config.showACL = true;
             this.render($container);
         })
         .on('render', function() {
--- a/src/destination/selector.js
+++ b/src/destination/selector.js
@@ -104,7 +104,7 @@ export default function destinationSelectorFactory($container, config) {
              * Get the current selected class uri
              * @returns {String} the selected uri
              */
-            var getSelectedUri = function getSelectedUri() {
+            const getSelectedUri = () => {
                 var select = self.resourceSelector.getSelection();
                 var uris;
                 //validate the selection
--- a/src/destination/selector.js
+++ b/src/destination/selector.js
@@ -113,7 +113,7 @@ export default function destinationSelectorFactory($container, config) {
             };
 
             var getSelectedACLTransferMode = function getSelectedACLTransferMode() {
-                return 'acl.keep.original';
+                return $('input[name="acl-mode"]:checked').val();
             };
 
             if (this.config.taskQueue) {
--- a/src/destination/selector.js
+++ b/src/destination/selector.js
@@ -31,6 +31,7 @@ import loadingButtonFactory from 'ui/loadingButton/loadingButton';
 import taskCreationButtonFactory from 'ui/taskQueueButton/standardButton';
 import selectorTpl from 'ui/destination/tpl/selector';
 import 'ui/destination/css/selector.css';
+import uuid from 'lib/uuid';
 
 var defaultConfig = {
     title: __('Copy to'),
@@ -90,6 +91,7 @@ export default function destinationSelectorFactory($container, config) {
     )
         .setTemplate(selectorTpl)
         .on('init', function () {
+            this.config.uniqId = uuid(8);
             this.render($container);
         })
         .on('render', function () {
--- a/src/destination/selector.js
+++ b/src/destination/selector.js
@@ -53,6 +53,8 @@ var defaultConfig = {
  * @param {Object} [config.taskQueue] - define the taskQueue model to be used (only useful if the triggered action uses the task queue)
  * @param {String} [config.taskCreationUrl] - the task creation endpoint (only required if the option taskQueue is defined)
  * @param {Object} [config.taskCreationData] - optionally define the data that will be sent to the task creation endpoint
+ * @param {Boolean} [config.showACL] - optionally define if ACL controls shall appear on the interface
+ * @param {Boolean} [config.ACLMode] - define ACL default behavior on the interface
  * @param {Function} [config.preventSelection] - prevent selection callback (@see ui/resource/selectable)
  * @returns {destinationSelector} the component itself
  */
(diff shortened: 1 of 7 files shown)


src/destination/selector.js
```diff
@@ -108,6 +112,10 @@ export default function destinationSelectorFactory($container, config) {
                 }
             };
 
+            var getSelectedACLTransferMode = function getSelectedACLTransferMode() {
+                return $('input[name="acl-mode"]:checked').val();
+            };
```
jsconan, line 117: **thought:** New code can be written in ES6 in this repo. Actually, when possible, we can also convert existing code to ES6. Anyway not a blocker.

**suggestion:**
```suggestion
            const getSelectedACLTransferMode = () => $('input[name="acl-mode"]:checked').val();
```

src/destination/tpl/selector.tpl
```diff
@@ -4,9 +4,18 @@
...
         {{/if}}
         <div class="selector-container"></div>
+        {{#if showACL}}
+        <div class="permissions-settings-container">
+            <p class="section-title">{{__ 'Permission settings'}}</p>
+            <input type="radio" id="acl-keep-original" {{#equal aclTransferMode 'acl.keep.original' }}checked {{/equal}}name="acl-mode" value="acl.keep.original" />
+            <label for="acl-keep-original">{{__ 'Keep the original permissions'}}</label><br>
+            <input type="radio" id="acl-use-destination" {{#equal aclTransferMode 'acl.use.destination' }}checked {{/equal}}name="acl-mode" value="acl.use.destination" />
+            <label for="acl-use-destination">{{__ 'Inherit the permission of the parent folder'}}</label>
+        </div>
```
jsconan, line 17: **issue:** There are severals issues in this code:
- The component does not uses a form despite it manages input fields. While this is ok since the component manage the inputs by itself, it may impact the container if it is also a form as the name fields will be part of it silently. Suggestion: don't use `name`, prefer `data-name` or similar.
- The labels are identified, but the ID is not guaranteed to be unique within the page, leading to a possible collision. Suggestion: don't use ID, prefer nested label instead `<label><input /></label>`, or use dynamic unique ID if the nested label cannot apply

jsconan (commented):
It looks nice. However, I have a few remarks. Please see below.

As a side note, could I suggest fixing the visual unit test too?
- you can add this to the file `test.html`, in the stylesheet:
    ```css
      div#qunit {
          position: relative;
      }
    ```
- you could also add the config for the visual test:
    ```javascript
            ...,
            showACL: true,
            aclTransferMode: 'acl.keep.original'
    ```

src/destination/selector.js
rgomeztao: updated, thanks!

src/destination/tpl/selector.tpl
rgomeztao: updated. Unfortunately, removing name from the inputs will remove the relation between the button set. however,  thats something we may fix with some javascript :)

src/destination/tpl/selector.tpl
```diff
@@ -4,9 +4,18 @@
...
+        <p class="section-title">{{description}}</p>
         {{/if}}
         <div class="selector-container"></div>
+        {{#if showACL}}
+        <div class="permissions-settings-container">
+            <p class="section-title">{{__ 'Permission settings'}}</p>
+            <input type="radio" class="acl-mode-radio acl-keep-original" {{#equal aclTransferMode 'acl.keep.original' }}checked {{/equal}} value="acl.keep.original" />
+            <label for="acl-keep-original">{{__ 'Keep the original permissions'}}</label><br>
+            <input type="radio" class="acl-mode-radio acl-use-destination" {{#equal aclTransferMode 'acl.use.destination' }}checked {{/equal}} value="acl.use.destination" />
+            <label for="acl-use-destination">{{__ 'Inherit the permission of the parent folder'}}</label>
```
jsconan, line 16: **issue:** The label is not linked with the input. Actually, the `for` attribute needs an ID.
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attributes

**suggestion:** Use dynamic ID generated from the component
```suggestion
            <input id="acl-keep-original{{uniqId}}" type="radio" class="acl-mode-radio acl-keep-original" {{#equal aclTransferMode 'acl.keep.original' }}checked {{/equal}} value="acl.keep.original" />
            <label for="acl-keep-original{{uniqId}}">{{__ 'Keep the original permissions'}}</label><br>
            <input id="acl-use-destination{{uniqId}}" type="radio" class="acl-mode-radio acl-use-destination" {{#equal aclTransferMode 'acl.use.destination' }}checked {{/equal}} value="acl.use.destination" />
            <label for="acl-use-destination{{uniqId}}">{{__ 'Inherit the permission of the parent folder'}}</label>
```

jsconan (changes requested):
Almost there, still a few things to address, unfortunately.

src/destination/selector.js
```diff
@@ -85,13 +89,17 @@ export default function destinationSelectorFactory($container, config) {
...
+        .on('init', function () {
             this.render($container);
         })
-        .on('render', function() {
+        .on('render', function () {
             var self = this;
             var $component = this.getElement();
 
+            $('.acl-mode-radio').change(function () {
+                $('.acl-mode-radio').not(this).prop('checked', false);
```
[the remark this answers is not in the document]
rgomeztao: fixed, ill keep in mind for the future :)

src/destination/selector.js
```diff
@@ -108,6 +116,8 @@ export default function destinationSelectorFactory($container, config) {
                 }
             };
 
+            const getSelectedACLTransferMode = () => $('input.acl-mode-radio:checked').val();
```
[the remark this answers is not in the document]
rgomeztao: fixed, thanks!

test/destination/selector/test.css
```diff
@@ -0,0 +1,3 @@
+div#qunit {
+    position: relative;
+}
```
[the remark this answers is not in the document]
rgomeztao: fixed, well i took inspiration from here https://github.com/oat-sa/tao-core-ui-fe/blob/e3bff77a61f9cac5345acf49c269434c42ec0b30/test/contextualPopup/test.css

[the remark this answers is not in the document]
jsconan: Ok got it. But it was incomplete because the CSS also need to be loaded, like here: https://github.com/oat-sa/tao-core-ui-fe/blob/e3bff77a61f9cac5345acf49c269434c42ec0b30/test/contextualPopup/test.js#L1

src/destination/selector.js
```diff
@@ -85,13 +89,17 @@ export default function destinationSelectorFactory($container, config) {
         defaultConfig
     )
         .setTemplate(selectorTpl)
-        .on('init', function() {
+        .on('init', function () {
             this.render($container);
```
jsconan, line 93: **suggestion:** the `uniqId` needs to be set in the config
```suggestion
            this.config.uniqId = uuid(); // need to add this too: `import uuid from 'lib/uuid';`
            this.render($container);
```

jsconan:
Please also fix the lint issue at this line: https://github.com/oat-sa/tao-core-ui-fe/pull/555/files#diff-5f3fae7476c943fe8d4adcb6ee604ea84f8bb79fb5750d087ba85bdbf7ed0172R107

You can fix it easily by removing the useless var declaration (the function is self declarative)

rgomeztao:
> 

fixed :thinking:

i have chnaged the var for a const, just to be consistent with the next line, and changed the function by an arrow function
 
and apparently yes, it seems that neither lintering or prettier have not been working as expected on this project, usually i run them from command line, but on these project both were giving false negatives report :thinking:

src/destination/selector.js
rgomeztao: > import uuid from 'lib/uuid';

fixed and added the import line to imports area :) thanks!

src/destination/tpl/selector.tpl
jsconan: Correct, the name allows grouping the buttons. Please ignore my previous comment, the name should be present.

jsconan (approved):
It works nicely. Good job.

Regarding the names of the radio buttons, you are right, they should be present. Please ignore my previous comment.

**Review Checklist**
- [x] New code respects code style rules
- [x] New code respects best practices
- [ ] ~~New code is not subject to concurrency issues (if applicable)~~
- [x] Commits are atomic

jsconan (approved):

PR #581, opened 2023-12-11
Feature/Auto-release

jsconan:
Add an auto-release script: when a pull request is merged to develop, auto-release the changes.

### DIFF
--- a/.github/workflows/release-package.yml
+++ b/.github/workflows/release-package.yml
@@ -1,0 +1,39 @@
+name: Release NPM package
+
+on:
+    pull_request:
+        branches:
+            - develop
+        types: [closed]
+
+jobs:
+    auto-release:
+        if: github.event.pull_request.merged == true
+        name: Automated package release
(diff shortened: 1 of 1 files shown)


gabrielfs7 (approved):
LGTM. Thanks!

erkobridee (approved):

taorafael (approved):
standard rules from other repos, lgtm

.github/workflows/release-package.yml
```diff
@@ -0,0 +1,39 @@
+name: Release NPM package
+
+on:
+    pull_request:
```
taorafael, line 4: if you add this, you can also trigger the release manually eventually, if you would be pointing, for example, to an integration branch instead of develop. :)

```suggestion
    workflow_dispatch:
    pull_request:
```

jsconan: Thx, This would be a good improvement for the future.
However, for the time being, I don’t think the tool would be compatible.

PR #606, opened 2024-11-08
Fix/aut 3923/textreader image figure support

olga-kulish:
Related to https://oat-sa.atlassian.net/browse/AUT-3923

### Changes
1. `mediaAlignment/helper`
  In _TextReader PCI_, `parent.contentModel && parent.contentModel === 'inlineStatic'` didn't pass because `parent.contentModel` was undefined. Because `widget.refresh` didn't run, something wasn't updated, and image align reverted to _Inline_.
  It seemed fine to always call `widget.refresh` instead of `active/sleep/active` for other cases (_A-block, ChoiceInteraction choice_), not only for _Prompt_. Only for prompts of _MathEntry/Likert PCI_, there was now "_QTI XML compilation error_" when there was `<figure>` in the prompt (figure wasn't allowed before, for same reason as in TextReader PCI), but it was solved by another change in https://github.com/oat-sa/extension-tao-itemqti/pull/2613 
  This code was added in https://github.com/oat-sa/tao-core-ui-fe/pull/528, and `refresh` was added at the last moment.

2. `FigureStateActive`
 _Image Caption textarea_ in Authoring _side panel_: it was completely collapsed for newly added images. Add some min-height to it. _Image Caption textarea_ appears after you wrap inline image.

### Test
On unit05. Related PRs:
- https://github.com/oat-sa/tao-core/pull/4125
- https://github.com/oat-sa/tao-core-ui-fe/pull/606
- https://github.com/oat-sa/extension-pcisample/pull/172
- https://github.com/oat-sa/extension-tao-itemqti/pull/2613

### DIFF
--- a/src/mediaEditor/plugins/mediaAlignment/helper.js
+++ b/src/mediaEditor/plugins/mediaAlignment/helper.js
@@ -13,7 +13,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2021-2023  (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2021-2024  (original work) Open Assessment Technologies SA;
  */
 import context from 'context';
 import _ from 'lodash';
@@ -21,29 +21,6 @@ import _ from 'lodash';
 export const FLOAT_LEFT_CLASS = 'wrap-left';
 export const FLOAT_RIGHT_CLASS = 'wrap-right';
 export const CENTER_CLASS = 'tao-centered';
-
-const searchRecurse = (parentElement, serial) => {
-    if (!parentElement) {
-        return null;
-    }
-    if (parentElement.serial === serial) {
-        return parentElement;
-    }
-    let found = null;
-    _.some(parentElement['elements'], childElement => {
-        if (childElement.serial === serial) {
-            found = parentElement;
-        } else if (childElement['elements']) {
-            found = searchRecurse(childElement, serial);
-        } else if (childElement['prompt']) {
-            found = searchRecurse(childElement.prompt.bdy, serial);
-        }
-        if (found) {
-            return true;
-        }
-    });
-    return found;
-};
 
 export const positionFloat = function positionFloat(widget, position) {
     if (!position) {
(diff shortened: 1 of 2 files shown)


src/mediaEditor/plugins/mediaAlignment/helper.js
```diff
@@ -22,29 +22,6 @@ export const FLOAT_LEFT_CLASS = 'wrap-left';
...
-            found = searchRecurse(childElement, serial);
-        } else if (childElement['prompt']) {
-            found = searchRecurse(childElement.prompt.bdy, serial);
-        }
-        if (found) {
-            return true;
-        }
-    });
-    return found;
-};
```
oatymart, line 46: Tagging @marpesia to see if she remembers something about the problem this code was added to solve 2 years ago.

marpesia: The code is working well on Authoring.
It rings a bell that it was done to be backwards compatible with previous items. This code was implemented at the end of 2022. Maybe it won't be an issue anymore.
I couldn't find any test exported from that time to validate my thoughts.

marpesia: I found the same code on itemqti. I think it is to replace any old item with the Figure tag. https://github.com/oat-sa/extension-tao-itemqti/pull/2279/commits/ba70db0af1f5e8c9cac665bf8a9fff45acdba601

Regarding the refresh() part, I remember we had some issues updating the CKEditor and keeping the changes.

marpesia (approved):

src/mediaEditor/plugins/mediaAlignment/helper.js
olga-kulish: Thanks! I checked it again and found at least one difference: in A-block, when `figure` is added in the middle of text inside `<p>`, the content of `<p>` is split to mutliple lines ('`<figure>` is not allowed inside `<p>`? And Prompt or TextReader use `<div>` not `<p>`). 
In my version, user wouldn't see that until he focuses text in A-block again, or reopens the item. 

Screenshot_21

I'll restore old version, and add TextReader PCI check. Not sure how to check for A-block. As for `contentModel`, both A-block and TextReader have `contentModel='itemBody'`. And `.qti-flow-container` in GapMatch/Hottext doesn't have this issue.

marpesia: Regarding the A-block, I tested on oat-dev to see differences, and the behaviour is the same. 
`<Figure>` is a block time, so it can't be inside a `<p>`.
I think it is ok.

PR #620, opened 2025-01-15
fix/ADF-1860/readonly-field-show-in-filters

shaveko:
related to https://oat-sa.atlassian.net/browse/ADF-1860

### Description

Show in the form list the field with the widget 

`http://www.tao.lu/datatypes/WidgetDefinitions.rdf#Readonly`

### How to test

Try to search an item when doing test authoring

`Equivalent Resource ID` search criteria should appear


jsconan (approved):

jsconan:
Quick code review. It seems straight forward. I didn't check though.

PR #646, opened 2025-05-01
Fix/TR-6452/vertical item layout improvements

olga-kulish:
Related to https://oat-sa.atlassian.net/browse/TR-6452
Related to https://oat-sa.atlassian.net/browse/TR-6577

[Deployed to unit01](https://oat-sa.atlassian.net/browse/TR-6452?focusedCommentId=305524)

Vertical writing:
- image positioning classes (`tao-centered`, `wrap-left`, `wrap-right`)
- underline should be to the right: support Safari < 18.2
- remove block-end margin from col in the last row (to allow full-height tao-overflow-y to fit all free space)
- margins for typography elements created in Authoring for A-block: p, ul, ol, h1-h6, blockquote, table
- `txt-combine-upright-all`: digits in interaction feedbacks should use `text-combine-upright: all`  (TR-6577)

### DIFF
--- a/scss/inc/_base.scss
+++ b/scss/inc/_base.scss
@@ -492,6 +492,13 @@ body {
         text-decoration: underline;
         font-size: inherit;
     }
+    .writing-mode-vertical-rl .txt-underline {
+        // should be 'text-underline-position: right', but for now to support Safari < 18.2:
+        // (box-shadow may also give space to ruby tags)
+        text-decoration: none;
+        box-shadow: -1px 0 0 -0.5px currentColor inset;
+        border-right: 1px solid currentColor;
+    }
 
     .txt-highlight {
         padding: 0 5px;
--- a/scss/inc/_base.scss
+++ b/scss/inc/_base.scss
@@ -1,7 +1,7 @@
 // this combines the most common features from normalize, h5bp main and tao base
 
 //these vars enables you to include this base in a scoped env.
-$mainContainer : html !default;
+$mainContainer: html !default;
 
 // Those styles can't be scoped because it's the base to use REM
 html {
@@ -29,17 +29,21 @@ body {
 }
 
 // cannot be scoped
-::-webkit-scrollbar-corner { background-color: ThreeDHighlight;}
+::-webkit-scrollbar-corner {
+    background-color: ThreeDHighlight;
+}
 
 #{$mainContainer} {
-
-    *, *:before, *:after {
+    *,
+    *:before,
+    *:after {
         -moz-box-sizing: border-box;
         -webkit-box-sizing: border-box;
         box-sizing: border-box;
     }
 
-    &, body{
+    &,
+    body {
         color: $textColor;
         font-family: $regularFont;
         font-weight: normal;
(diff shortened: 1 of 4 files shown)


scss/inc/_base.scss
```diff
@@ -420,12 +497,20 @@ body {
...
-        font-size : inherit;
+        font-size: inherit;
+    }
+    .writing-mode-vertical-rl .txt-underline {
+        text-underline-position: right;
+    }
+    .writing-mode-vertical-rl[data-useragent-browser='safari'] .txt-underline {
+        // for Safari < 18.2 which doesn't support "text-underline-position: right"
+        text-decoration: none;
+        border-right: 1px solid currentColor;
```
olga-kulish, line 508: I'm not sure if `[data-useragent-browser='safari']` is really needed, we can always use `border-right: 1px solid currentColor;` instead of underline, in all browsers.

Just when testing I found that in Chrome, on some zooms/resolutions, border seems to be 1px farther to the right than underline, and so it overlaps ruby tags making them less readable. Safari seems fine.

scss/inc/_grid.scss
```diff
@@ -30,16 +32,18 @@
         margin-block-end: 12px;
     }
     .alpha {
-        margin-left: 0 !important
+        margin-left: 0 !important;
```
oatymart, line 35: Not sure what this element is, does it need to become `margin-inline-start`?
Also isn't it written 12 times into the CSS unnecessarily?

scss/inc/_base.scss
oatymart: I don't have a preference. The extra attribute and coupling is a bit annoying, and it could be years before Safari 18.2 is the minimum supported browser, so if you want to change it for the `border-right` for all browsers it makes sense.

Also noticed that the `text-underline-position: right` disappears when the word is selected in Chrome, but the `border-right` does not.

scss/inc/_base.scss
```diff
@@ -420,12 +497,24 @@ body {
...
+    }
+
+    .txt-combine-upright-all {
```
olga-kulish, line 511: Is this class name ok with you?

Customers also want to use this in the authored text, so I'll suggest them to use this class. (or they can use item stylesheets; they agreed to edit html using CkEditor SourceDialog plugin)

oatymart: It's fine! (To be honest I had no idea what it meant, until I looked up [text-combine-upright](https://developer.mozilla.org/en-US/docs/Web/CSS/text-combine-upright).) I think no-one will ever use it but Japanese or Chinese authors, so a really specific name is fine.

scss/inc/_grid.scss
olga-kulish: Also don't know what it is, but let's change it.

Usually I'd prefer to change only minimal amount of styles, because this also affects RTL. (But maybe nobody uses RTL with CurretGen anymore)

scss/inc/_base.scss
olga-kulish: Interesting thing with selection!
It's difficult to choose, but I guess I'll leave `underline` because of ruby. It's just 1px, but they are already hard to read...

oatymart (approved):
Very nice job!

- [x] Pull request's target is not `master` (or `main`) - or is a release
- [ ] ~~Bundled code is present if needed~~
- [ ] ~~Version bump is applied if needed (packages AND root)~~
- [x] Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] Commits are atomic
- [ ] ~~Changelog is updated according to changes (if applicable)~~

PR #700, opened 2026-03-31
chore: github workflow: publish from tag

olga-kulish:
**For backports**:
After release tag and version bump are done manually,
Use github workflow to do `npm publish`
- to avoid running `npm run build` on the local machine
- to avoid having token with publishing permissions on the local machine

Based on https://github.com/oat-sa/tao-deliver-fe/blob/main/.github/workflows/publish-npm-released.yml

## Summary by CodeRabbit

* **Chores**
  * Added a manual workflow to publish tagged releases to NPM from a specified tag, with semantic version validation, optional release tag selection (defaults to "latest"), and secure token-based authentication.


kilatib (approved):
Looks safe fro me

